PR #3, opened 2019-05-09
Address many Itay questions and cleanup.

Bronson-Brown-deVost:
@zmbq , I worked through a several issues in this pull.  It still needs more cleanup and I still have some questions from you that I should answer.

I left the GroupJoin you didn't care for, but I did annotate it.  I can still change that to something less functional, and even more likely I will perform the underlying queries differently (more efficiently).

I fixed the return of the artefact change mutations in accordance with our discussion yesterday, so you now receive back the ArtefactDTO for the updated artefact as you want (it only sends back the polygon mask if your request updated the mask).

Feel free to inline any questions in the pull request.


zmbq (commented):
I went over the code and left some remarks.
I think my biggest issue is the breach of separation of concerns. It happened once when you had the repository parse URLs. It happened a second time in a less obvious way, but I think it's still important. We have a very clear (albeit a bit more involved) definition of which layer handles which classes, like so:

Controllers only know DTOs. They pass DTOs to services.
Services only have DTOs on their interfaces. They talk to repositories with data models. Sevices provide static functions that convert models to DTOs (we may move this to another entity type).
Repositories only have data models in their interfaces.
Repositories can use ....Query.Result objects internally for performing queries, but they should always convert them to DataModels when returning results.

data-access/EditionRepository.cs
```diff
@@ -143,15 +143,17 @@ public async Task ChangeEditionNameAsync(uint editionId, string name, UserInfo u
                     
                     // Now TrackMutation will insert the data, make all relevant changes to the owner tables and take
                     // care of main_action and single_action.
-                    var response = await _databaseWriter.WriteToDatabaseAsync(editionId, user, new List<MutationRequest>() { nameChangeRequest });
+                    var response = await _databaseWriter.WriteToDatabaseAsync(user, new List<MutationRequest>() { nameChangeRequest });
                     // Bronson: Where isn't there a response?
+                    // Itay: You get no response when you try to change to the name that is already being used.
+                    // This can happen when, after the database write begins, someone else changed it to the same name first.
```
zmbq: I think these are two cases:

1. The user hasn't specified a new name. This can be either swallowed by the server or return a 400 Bad Request.
2. The user has specified a name that's used by another edition. In this case return a 409 Conflict. We may add a check for this in the frontend, as we have all the user's editions anyway. The server sholudn't rely on frontend checks, though, and should always perform them itself.

data-access/Models/UserModels.cs
```diff
@@ -51,7 +51,14 @@ public UserInfo(string requestPath, IUserRepository userRepository)
...
                     _editionId = uint.TryParse(scrollEditionMatch.Groups[1].Value, out var i) ? i : 0;
-                else _editionId = 0;
+                else
+                {
+                    scrollEditionTest = new Regex(@"^.*/editions/(\d{1,32}).*$");
+                    scrollEditionMatch = scrollEditionTest.Match(_requestPath);
+                    if (scrollEditionMatch.Groups.Count == 2)
+                        _editionId = uint.TryParse(scrollEditionMatch.Groups[1].Value, out var i) ? i : 0;
+                    else _editionId = 0;
+                }
```
zmbq: Is _requestPath the URL passed to the controller? The model shouldn't parse it in any way. If we change the URL (drop the /editions, for example, or rename /editions to /edition), the data model shouldn't care. If you need the edition ID, add is a a parameter, and let the caller parse the URL.

sqe-http-api/Controllers/ArtefactController.cs
```diff
@@ -24,20 +22,48 @@ public ArtefactController(IUserService userService, IArtefactService artefactSer
             this._userService = userService;
         }
         
+        private void ParseOptionals(List<string> optionals, out bool images, out bool masks)
+        {
+            images = masks = false;
+            if (optionals == null) return;
```
zmbq: This is very petty, but I prefer to put return in a separate line...
Also, I want to see what's the difference between Githubs "Add Single command" and "Start a review"

sqe-http-api/Controllers/ArtefactController.cs
```diff
@@ -24,20 +22,48 @@ public ArtefactController(IUserService userService, IArtefactService artefactSer
...
+        
         /// <summary>
         /// Provides a listing of all artefacts that are part of the specified edition
         /// </summary>
         /// <param Name="editionId">Unique Id of the desired edition</param>
         /// <param Name="images">Defines whether image references should be included in the response</param>
         [AllowAnonymous]
         [HttpGet("edition/{editionId}/artefact/list")]
-        public async Task<ActionResult<ArtefactListDTO>> GetArtefacts([FromRoute] uint editionId, [FromQuery] string images = "false")
+        public async Task<ActionResult<ArtefactListDTO>> GetArtefacts([FromRoute] uint editionId, [FromQuery] List<string> optional = null)
```
zmbq: Actually, there's no point in specifying a null default value. ASP.NET Core's default binder passes an empty list by default. At least that's what happened when I tried it.

sqe-http-api/Controllers/ArtefactController.cs
```diff
@@ -78,33 +104,27 @@ public async Task<ActionResult<ArtefactListDTO>> GetArtefacts([FromRoute] uint e
...
         }
 
-        public class UpdateArtefactMask
-        {
-            public string mask { get; set; }
-        }
-
         public class UpdateArtefactDTO
         {
             public string mask { get; set; }
```
zmbq, line 130: Isn't it possible to update the Artefact's name, too?

sqe-http-api/DTOs/Artefact.cs
```diff
@@ -11,16 +13,46 @@ public class ArtefactDTO
...
+
+                imagedObjectId = ImagedObjectIdFormat.Serialize(artefact.institution, artefact.catalog_number_1, artefact.catalog_number_2),
+
+                name = artefact.name,
+                side = artefact.catalog_side == 0 ? ArtefactDTO.ArtefactSide.recto : ArtefactDTO.ArtefactSide.verso, 
+                zOrder = 0,
+            };
+        }
+        
+        public static ArtefactListDTO QueryArtefactListToArtefactListDTO(List<ArtefactsOfEditionQuery.Result> artefacts, uint editionId)
```
zmbq: First, I think we placed all the ModelToDTO methods in services and not the DTOs themselves. I think the DTOs should n't be aware of the models.
Second, the repositories should return models, not query result objects

sqe-http-api/DTOs/ImagedObject.cs
```diff
@@ -24,4 +25,42 @@ public class ImagedObjectListDTO
...
+        public static string Serialize(string institution, string catalogNumber1, string catalogNumber2)
+        {
+            return institution + "-"
+                               + catalogNumber1
+                               + (string.IsNullOrEmpty(catalogNumber2)
+                                   ? ""
+                                   : "-" + catalogNumber2);
+        }
+
+        // Warning: I like the idea to create a readable imagedObjectID from the object's cataloguing information.
```
zmbq, line 43: We only created these mock IDs because the database didn't have them. Can't we just have the database handle them (even making them numeric)?

sqe-http-api/DTOs/ImagedObject.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using Microsoft.EntityFrameworkCore.Storage.ValueConversion;
```
zmbq: What is this for?

sqe-http-api/Services/ArtefactService.cs
```diff
@@ -2,6 +2,7 @@
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
+using Google.Protobuf.WellKnownTypes;
```
zmbq: What's this for? Why do we even use Protobufs?

sqe-http-api/Services/ImagedObjectService.cs
```diff
@@ -71,46 +70,30 @@ public async Task<ImagedObjectListDTO> GetImagedObjectsAsync(uint? userId, uint
...
+            var artefacts = ArtefactDTOTransform.QueryArtefactListToArtefactListDTO(
+                (await _artefactRepository.GetEditionArtefactListAsync(userId, editionId, withMasks)).ToList(), 
+                editionId
+            );
 
             // Bronson: This is *really* unclear, GroupJoin is not something people usually understand.
             // A few comments are warranted here, or even better - dropping LINQ altogether here
             // I spent a couple of minutes reading GroupJoin's documenation, and I'm not closer to figuring out 
             // what is being grouped.
             result.imagedObjects = result.imagedObjects.GroupJoin(
```
zmbq, line 88: I'm still not clear on what this does. I think it's best to describe with a short example in the comments.

zmbq (changes requested):

sqe-http-api/Controllers/ArtefactController.cs
Bronson-Brown-deVost: Thanks, you must be right.  I have changed all 3 occurences.

sqe-http-api/Controllers/ArtefactController.cs
Bronson-Brown-deVost: Fine with me.  Changed 6 instances.

sqe-http-api/Controllers/ArtefactController.cs
Bronson-Brown-deVost: Yep, in the UpdateArtefactDTO you pass to the PUT request you can set any combination of the three: "mask", "name", and "position".

sqe-http-api/DTOs/ImagedObject.cs
Bronson-Brown-deVost: I can just give you the image_catalog_id, that is the unique identifier for an institution-plate-fragment-side.

zmbq: So our ImagedObject id will be the image_catalog_id? This seems very clean and neat, yes. So we should change the ImagedObjectDTO's id field to a number?

sqe-http-api/DTOs/ImagedObject.cs
Bronson-Brown-deVost: My IDE tells me that of all those usings, the only one used by the code in that namespace is: using System.Collections.Generic;

I have gone through and cleaned out the unused "usings" (~50 occurences).

sqe-http-api/Services/ArtefactService.cs
Bronson-Brown-deVost: I have absolutely no idea...  It isn't used by anything, so I removed it.

sqe-http-api/Services/ImagedObjectService.cs
Bronson-Brown-deVost: I can do that.  I am also willing to drop Linq here.  I assume that this "functional" approach is optimized, but that may be an invalid assumption.
Basically I grab all the images requested (without artefact data), and I separately grab all the artefact data requested (I should have done that concurrently).  Then this code gathers the artefacts and inserts the proper ones into their corresponding image objects (so the artefacts in those image objects are no longer null).

data-access/EditionRepository.cs
Bronson-Brown-deVost: 1. Agreed, If the user hasn't specified a name, we should return immediately with a 404. 

2. This is not the case I had in mind, there is no restriction on naming and multiple edition are allowed to have the same name, so there is never a 409 Conflict possible in this case.  The difference is between whether a change was made to the database or no change was made to the database (hence response.Count == 0).  If you try to change the name to what the edition is already named, no data will be written to the database.  In such a case, should we inform the user that the request did nothing (though it was otherwise valid, and the database does reflect the desired outcome)?  Or should we just tell them it was successful (since the state of the database is exactly what they requested)?  Other ideas?

data-access/Models/UserModels.cs
Bronson-Brown-deVost: Yes, this already caused me trouble once.  I am fixing it now as you suggest.

Bronson-Brown-deVost: Ok. It is finished now. Thanks.

sqe-http-api/DTOs/Artefact.cs
Bronson-Brown-deVost: That is fine with me.  This particular code is used by both the ImagedObjectService and the ArtefactService, so I guess it needs to live outside those.  I created a file ArtefactDTOTransformService.cs, but perhaps there is a better place to put such multi-use transformers.

I have to run to a parent-teacher meeting right now.  Can we talk on slack about the issue of these intermediary models between the repositories and the services.

sqe-http-api/DTOs/ImagedObject.cs
Bronson-Brown-deVost: Yes.  Shall I make the change now, or would you rather we do it in a separate branch+pull request.

data-access/EditionRepository.cs
zmbq: Do you mean I can have two different editions with exactly the same name? Isn't that going to be a problem for users? How am I going to tell the difference between the two? Obviously two users can have two editions with the same name, but one user with duplicate names? I think this is mostly a source for trouble. "I changed this and that, but I can't see these changes!" emails. Frantic emails.

As for returning an error when nothing changed as opposed to returning 200 - let's look at it from the client's perspective. I think it's better not to report an error at all - just like an UPDATE statement that doesn't change anything doesn't fail, or "x = x" doesn't throw an exception. However, this is a tiny issue, and if we return an error - that's fine too.

sqe-http-api/DTOs/Artefact.cs
zmbq: We have a bunch of conversions like this, let's all put them in the same place. Note that this is *not* a service, but rather a static class (a class that only has static methods). It is not injectable, and isn't registered anywhere.

sqe-http-api/DTOs/ImagedObject.cs
zmbq: Do it now. We'll fix the frontend's DTO. It will keep working, anyway, since the DTO expects a string, and all numbers are automatically strings in Javascript, too.

sqe-http-api/Services/ImagedObjectService.cs
zmbq: I think I prefer a comment, and keeping the functional approach. It'll force us (well, me, right now) to think more functionally. Don't worry about performance here, we will have zero CPU issues in the backend.

data-access/EditionRepository.cs
Bronson-Brown-deVost: Ok, I can see the value of 200 here, so maybe we just do that.  You bring up a good point when a user has more than one edition with the same name, that can be confusing (though the database technically allows it, since a "power user" could look at the editionId to differentiate).  I too think that we should prevent the average user from having two editions with the same name, you are right that they will probably confuse things.  So my question is, do we use the ScrollEditor website logic to prevent such an action entirely (you should already have the info there to know this case), do we use ScrollEditor website logic to provide a strong warning (with a note to look at the editionId), but ultimately allow the user to do it, or do we prevent this operation in the API (you would also probably block it in the website logic too)? I will email this question to Ingo too for his opinion.

sqe-http-api/DTOs/Artefact.cs
Bronson-Brown-deVost: Yes, you are absolutely right, we will have many of these class "transformers", which will all be static (some may even build upon others with inheritance). So should I put it in the "Helpers" folder, or should I create a new folder (and namespace) with these static class transformation methods?  What would I call it?

sqe-http-api/DTOs/ImagedObject.cs
Bronson-Brown-deVost: One little problem is that for each ImagedObjectDTO there are actually two image_catalog_id numbers (one for the recto and one for the verso).  I could just join those two with a hyphen "987-10267" and we would not have the decomposition problem with the other system.  What do you think?

sqe-http-api/DTOs/Artefact.cs
zmbq: I think the Helpers folder is great for this.

sqe-http-api/DTOs/ImagedObject.cs
zmbq: I think this brings us back into the parsing IDs. And although now IDs can't be misparsed (there are no hyphens in numbers), it's still a bit of a pain. Can you use the recto image ID?

Bronson-Brown-deVost: Yeah, I can do that.  

Perhaps we could make the object a bit more explicit so that it communicates sides better?  The endpoint for the imagedObject could default to recto+verso (and we give an optional for explicit request of recto or verso [and recto+verso?]).  The object you get back would have the image_catalog_id of the requested side as its id (default to recto), would have a property telling you the side (default to recto), and could also give you an oppositeSideId with the image_catalog_id of the opposite side (default to verso). Maybe this could be helpful as we build out better recto/verso support?  It would be the same DTO we are using now just with an added "side" property.

sqe-http-api/Services/ImagedObjectService.cs
Bronson-Brown-deVost: Done.

sqe-http-api/DTOs/Artefact.cs
Bronson-Brown-deVost: Sounds good to me.  I will do it today.

sqe-http-api/DTOs/ImagedObject.cs
zmbq: I actually don't remember what we decided on recto/verso and images. Do we treat the verso as an entirely different image from the recto? If so, these are two separate ImagedObjects, each with its own side of course. And a link to the other one (otherSideId: ...)

sqe-http-api/Services/ImagedObjectService.cs
zmbq: OK, but now there's a problem with the github interface - I don't see your new changes in this pull request...

sqe-http-api/DTOs/ImagedObject.cs
Bronson-Brown-deVost: Sorry for the delay.  I am glad we have gone through this, and now I think I have come back to the original suggestion of using an id like IAA-994-1, which represents the object as it was designated by the museum.  The object IAA-994-1, for example, has recto and verso images (maybe the system somehow accommodates even more than that someday).  

So, let's leave the imagedObjectDTO as it is, I can add the necessary logic in the API/database to be able to search via this id (but avoid the composition/decomposition issues).  I think I originally liked what you did with that designation because it fits what the API user expects: a designation of an object, not of an image (or image series).

sqe-http-api/Services/ImagedObjectService.cs
Bronson-Brown-deVost: At the bottom of this page you see the commits that I have added since the review started.  I will continue making commits to this branch as I make the changes we have discussed here.  Then you just pull the request and everything will be done.

sqe-http-api/DTOs/ImagedObject.cs
zmbq: In this case, I think we should just add this field to the database, shouldn't we?

data-access/Models/UserModels.cs
```diff
@@ -25,52 +22,31 @@ public class UserEditionPermissions
...
         private readonly IUserRepository _userRepo;
-        private uint? _editionId;
+        public uint? editionId;
         private uint? _editionEditorId;
         private bool? _mayWrite;
         private bool? _mayLock;
         private bool? _isAdmin;
 
-        public UserInfo(string requestPath, IUserRepository userRepository)
+        public UserInfo(uint? userId, uint? editionId, IUserRepository userRepository)
```
zmbq, line 32: I understand the reason for the UserInfo class. It may better be called CurrentUserContext. I am certain it is not a DataModel, though - it's not mapped to the database. I guess it should be moved elsewhere within the data access project.

zmbq, line 32: I'm not crazy about IUserRepository being passed to this class. It should be injected. However, I understand why you need it, and I don't think making the UserInfo class injectable is very helpful. So I suggest we keep it like this, and just add a comment saying something that's we're not really sure this is the right solution, but it's working in is good enough for now. You know, just to avoid someone going over the code a couple of years from now, thinking we were all jerks.

sqe-http-api/Controllers/ImagedObjectController.cs
```diff
@@ -27,9 +27,11 @@ public ImagedObjectController(IUserService userService, IImagedObjectService ima
         private void ParseOptionals(List<string> optionals, out bool artefacts, out bool masks)
         {
             artefacts = masks = false;
-            if (optionals == null) return;
+            if (optionals == null) 
+                return;
             artefacts = optionals.Contains("artefacts");
-            if (!optionals.Contains("masks")) return;
+            if (!optionals.Contains("masks"))
```
zmbq, line 33: The following is a very petty comment:
I think it's better to write:
```
masks = optionals.Contains("masks");
artefacts = masks || optionals.Contains("artefacts");
```

It's clearer.

sqe-http-api/DTOs/ImagedObject.cs
Bronson-Brown-deVost: Yep, I will add it to the database.  Thanks for talking through this.

data-access/EditionRepository.cs
Bronson-Brown-deVost: I removed the superfluous check.  We should work in conjunction with Ingo to fully resolve this issue in a later PR.

sqe-http-api/DTOs/ImagedObject.cs
Bronson-Brown-deVost: Done

Bronson-Brown-deVost:
Ok.  Sorry it took so long, I had a lot going on in between.  The PR is ready now.  

We will push the issue of uniqueness in edition names to a later PR; that will require a bit more discussion and possible changes to the database as well.

Note that this update requires v0.9.4 or higher of the SQE_Database (the dockerfiles here have been updated accordingly).

zmbq (approved):

PR #5, opened 2019-05-15
Fully support imaged object id

Bronson-Brown-deVost:
I saw that I failed to fully implement the imaged object id.  It is fixed now along with some bugfixes on the route names for artefacts.

### DIFF
--- a/data-access/Models/ArtefactModels.cs
+++ b/data-access/Models/ArtefactModels.cs
@@ -7,10 +7,8 @@
         public string Mask { get; set; }
         public string TransformMatrix { get; set; }
         public short ZIndex { get; set; }
-        public string Institution { get; set; }
-        public string CatalogNumber1 { get; set; }
-        public string CatalogNumber2 { get; set; }
         public byte CatalogSide { get; set; }
         public uint ImageCatalogId { get; set; }
+        public string ImagedObjectId { get; set; }
     }
 }
--- a/sqe-http-api/Controllers/ArtefactController.cs
+++ b/sqe-http-api/Controllers/ArtefactController.cs
@@ -57,7 +57,7 @@ namespace SQE.SqeHttpApi.Server.Controllers
         /// <param Name="ArtefactId">Unique Id of the desired artefact</param>
         /// <param Name="images">Defines whether image references should be included in the response</param>
         [AllowAnonymous]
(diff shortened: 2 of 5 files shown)


data-access/Models/ArtefactModels.cs
```diff
@@ -7,10 +7,8 @@ public class ArtefactModel
         public string Mask { get; set; }
         public string TransformMatrix { get; set; }
         public short ZIndex { get; set; }
```
Bronson-Brown-deVost, line 9: We don't need Institution, CatalogNumber1, or CatalogNumber2 in the artefact model, the ImagedObjectId will suffice.

sqe-http-api/Controllers/ArtefactController.cs
```diff
@@ -57,7 +57,7 @@ public async Task<ActionResult<ArtefactListDTO>> GetArtefacts([FromRoute] uint e
         /// <param Name="ArtefactId">Unique Id of the desired artefact</param>
         /// <param Name="images">Defines whether image references should be included in the response</param>
         [AllowAnonymous]
-        [HttpGet("editions/{editionId}/artefacts/{ArtefactId}")]
+        [HttpGet("editions/{editionId}/artefacts/{artefactId}")]
```
Bronson-Brown-deVost, line 60: Somehow the route parameter artefactId was changed to ArtefactId by my fixing of the classes to use PascalCase.  I fix it here.

sqe-http-api/DTOs/ImagedObject.cs
```diff
@@ -20,43 +20,5 @@ public class ImagedObjectDTO
     public class ImagedObjectListDTO
     {
         public List<ImagedObjectDTO> imagedObjects { get; set; }
-    };
-
-    public static class ImagedObjectIdFormat
```
Bronson-Brown-deVost, line 25: As agreed, we no longer need this function, and it should not have been in the DTO anyway.  It is removed now.

data-access/Models/ArtefactModels.cs
zmbq: Yes, you are correct. Can you please open an issue to Talya show she removes them from the frontend DTO?

Bronson-Brown-deVost: No problem, I just did it.  You can merge this PR now.

zmbq (approved):

PR #6, opened 2019-05-16
User management

Bronson-Brown-deVost:
The user management endpoints are written and tested.

Part of the user management uses an emailer.  I use Gmail for this right now until we know what will be used long term.  For that functionality to work, you will need to provide the email credentials as environment variables.  Set your gmail username with SQE_GMAIL_USERNAME (i.e., the name part of name@gmail.com), and your password with SQE_GMAIL_PASSWORD (probably you can set these environment variables in visual studio).  You must also make sure your gmail account allows less secure apps (https://support.google.com/accounts/answer/6010255?hl=en).

### DIFF
--- a/data-access/Models/ScrollModels.cs
+++ b/data-access/Models/ScrollModels.cs
@@ -12,7 +12,7 @@ namespace SQE.SqeHttpApi.DataAccess.Models
         public bool Locked { get; set; }
         public bool IsPublic {get; set;}
         public DateTime? LastEdit { get; set; }
-        public User Owner { get; set; }
+        public UserToken Owner { get; set; }
     }
 
     public class ScrollName
@@ -32,7 +32,7 @@ namespace SQE.SqeHttpApi.DataAccess.Models
 
     public class Share
     {
-        public User User { get; set; }
+        public UserToken UserToken { get; set; }
         public Permission Permission { get; set; }
     }
 
--- a/data-access/Models/UserModels.cs
+++ b/data-access/Models/UserModels.cs
@@ -7,7 +7,18 @@ namespace SQE.SqeHttpApi.DataAccess.Models
     {
         public string UserName { get; set; }
         public uint UserId { get; set; }
+    }
+    
+    public class UserToken : User
+    {
         public string Token { get; set; }
+    }
+
+    public class UserEmail : User
+    {
+        public string Forename { get; set; }
+        public string Surname { get; set; }
+        public string Email { get; set; }
     }
     
     public class UserEditionPermissions
--- a/data-access/Models/UserModels.cs
+++ b/data-access/Models/UserModels.cs
@@ -9,9 +9,23 @@ namespace SQE.SqeHttpApi.DataAccess.Models
         public uint UserId { get; set; }
     }
     
+    public class UserLoginResponse : User
+    {
+        public byte Activated { get; set; }
+        public string Email { get; set; }
+    }
+    
     public class UserToken : User
     {
         public string Token { get; set; }
+    }
+
+    public class DetailedUserToken : UserToken
+    {
+        public string Forename { get; set; }
+        public string Surname { get; set; }
+        public string Organization { get; set; }
+        public string Email { get; set; }
     }
 
     public class UserEmail : User
--- a/data-access/Models/UserModels.cs
+++ b/data-access/Models/UserModels.cs
@@ -11,7 +11,7 @@ namespace SQE.SqeHttpApi.DataAccess.Models
     
     public class UserLoginResponse : User
     {
-        public byte Activated { get; set; }
+        public bool Activated { get; set; }
         public string Email { get; set; }
     }
     
@@ -20,7 +20,7 @@ namespace SQE.SqeHttpApi.DataAccess.Models
         public string Token { get; set; }
     }
 
-    public class DetailedUserToken : UserToken
+    public class DetailedUserWithToken : UserToken
     {
         public string Forename { get; set; }
         public string Surname { get; set; }
--- a/data-access/Queries/User.cs
+++ b/data-access/Queries/User.cs
@@ -2,14 +2,14 @@
 
 namespace SQE.SqeHttpApi.DataAccess.Queries
 {
-    internal class UserQueryResponse : IQueryResponse<User>
+    internal class UserQueryResponse : IQueryResponse<UserToken>
     {
         public string user_name { get; set; }
         public uint user_id { get; set; }
 
-        public User CreateModel()
+        public UserToken CreateModel()
         {
-            return new User
+            return new UserToken
             {
                 UserName = user_name,
                 UserId = user_id
@@ -42,38 +42,29 @@ SELECT edition_editor_id AS EditionEditionEditorId,
        is_admin AS IsAdmin
 FROM edition_editor
 WHERE edition_id = @EditionId AND user_id = @UserId";
-
-        public class Return
-        {
-            private uint edition_editor_id { get; set; }
-            private bool may_write { get; set; }
-            private bool may_lock { get; set; }
-            private bool may_read { get; set; }
-            private bool is_admin { get; set; }
-        }
     }
 
     /// <summary>
-    /// Checks whether an account with @Username or @Email exists and whether it is authenticated yet.
+    /// Checks whether an account with @Username or @Email exists and whether it is activated yet.
     /// </summary>
-    internal static class CheckUserAuthentication
+    internal static class CheckUserActivation
     {
         public const string GetQuery = @"
-SELECT user_id, authenticated, user_name, email
+SELECT user_id, activated, user_name, email
 FROM user
 WHERE user_name = @Username OR email = @Email";
 
-        public class Return
+        public class Result
         {
             public int user_id { get; set; } 
-            public bool authenticated { get; set; } 
+            public bool activated { get; set; } 
             public string user_name { get; set; }
             public string email { get; set; }
         }
     }
     
     /// <summary>
-    /// Creates a new unauthenticated user account for @Username, @Email, and @Password;
+    /// Creates a new user account for @Username, @Email, and @Password (the account is not activated);
     /// the fields @Forename, @Surname, and @Organization may be empty.
     /// </summary>
     internal static class CreateNewUserQuery
@@ -84,13 +75,13 @@ VALUES(@Username, @Email, SHA2(@Password, 224), @Forename, @Surname, @Organizati
     }
 
     /// <summary>
-    /// Deletes the user with id @UserId.  We can only ever delete accounts that have not been authenticated
+    /// Deletes the user with id @UserId.  We can only ever delete accounts that have not been activated
     /// (thus they never were able to create any data of their own).
     /// </summary>
     internal static class DeleteUserQuery
     {
         public const string GetQuery = @"
-DELETE FROM user WHERE user_id = @UserId AND authenticated = 0";
+DELETE FROM user WHERE user_id = @UserId AND activated = 0";
     }
 
     /// <summary>
@@ -103,22 +94,22 @@ UPDATE user
 SET pw = SHA2(@NewPassword, 224)
 WHERE user_id = @UserId
     AND pw = SHA2(@OldPassword, 224)
-    AND authenticated = 1 ## Only authenticated users may change their password.
+    AND activated = 1 ## Only activated users may change their password.
 ";
     }
 
     /// <summary>
-    /// Sets the user record to authenticated when provided with a valid token @Token.
+    /// Sets the user record to activated when provided with a valid token @Token.
     /// </summary>
     internal static class ConfirmNewUserAccount
     {
         public const string GetQuery = @"
 UPDATE user
 JOIN user_email_token USING(user_id)
-SET authenticated = 1
+SET activated = 1
 WHERE user_email_token.token = @Token 
     AND user_email_token.type = 'ACTIVATE_ACCOUNT' 
-    AND user.authenticated = 0 ## Only new users can authenticate an account
+    AND user.activated = 0 ## Only new users can activate an account
 ";
     }
     
@@ -130,7 +121,8 @@ WHERE user_email_token.token = @Token
     {
         public const string GetQuery = @"
 INSERT INTO user_email_token (user_id, token, type)
-VALUES(@UserId, @Token, @Type)";
+VALUES(@UserId, @Token, @Type)
+ON DUPLICATE KEY UPDATE token = @Token, date_created = NOW()";
 
         public const string Activate = "ACTIVATE_ACCOUNT";
         public const string ResetPassword = "RESET_PASSWORD";
@@ -161,6 +153,19 @@ WHERE email = @Email
     }
     
     /// <summary>
+    /// Returns the user info for the account with the user email token @Token.  Only use this if you know what you are doing!
+    /// </summary>
+    internal static class UserByTokenQuery
+    {
+        public const string GetQuery = @"
+SELECT user_name AS UserName, forename AS Forename, surname AS Surname, email AS Email
+FROM user
+JOIN user_email_token USING(user_id)
+WHERE user_email_token.token = @Token
+";
+    }
+    
+    /// <summary>
     /// Changes the password to @Password for the user who received the reset password token @Token.
     /// </summary>
     internal static class UpdatePasswordByToken
@@ -171,20 +176,7 @@ JOIN user_email_token USING(user_id)
 SET user.pw = SHA2(@Password, 224)
 WHERE user_email_token.token = @Token 
     AND user_email_token.type = 'RESET_PASSWORD' 
-    AND user.authenticated = 1 ## Only authenticated users can reset their password
-";
-    }
-    
-    /// <summary>
-    /// This is used to keep the user_email_token table clean and up-to-date.
-    /// It deletes records that are more than @Days old.
-    /// I could do this as a trigger in the database, and perhaps that would be better.
-    /// TODO: Decide whether to keep this or to use a trigger. Either way remove `date_expires` from the table.
-    /// </summary>
-    internal static class DeleteOldTokens
-    {
-        public const string GetQuery = @"
-DELETE FROM  user_email_token WHERE NOW() > date_created + interval @Days day
+    AND user.activated = 1 ## Only activated users can reset their password
 ";
     }
 }
--- a/data-access/Queries/User.cs
+++ b/data-access/Queries/User.cs
@@ -1,20 +1,53 @@
-﻿using SQE.SqeHttpApi.DataAccess.Models;
+﻿using System.Collections.Generic;
+using System.Linq;
+using SQE.SqeHttpApi.DataAccess.Helpers;
 
 namespace SQE.SqeHttpApi.DataAccess.Queries
 {
-    internal class UserQueryResponse : IQueryResponse<UserToken>
+    internal static class LoginQuery
     {
-        public string user_name { get; set; }
-        public uint user_id { get; set; }
+        public const string GetQuery = @"
+SELECT user_name AS UserName, user_id AS UserId, email AS Email, activated AS Activated
+FROM user 
+WHERE user_name = @UserName AND pw = SHA2(@Password, 224)
+";
+    }
 
-        public UserToken CreateModel()
+    /// <summary>
+    /// An extensible User info query builder.
+    /// TODO: Upgrade all user queries to use this.
+    /// </summary>
+    internal static class UserDetails
+    {
+        private const string _query = @"
+SELECT $Columns
+FROM user
+WHERE $Where";
+
+        /// <summary>
+        /// Returns a formatted query on the user table
+        /// </summary>
+        /// <param name="columns">Names of the columns to be retrieved (in snake_case)</param>
+        /// <param name="where">Names of the where parameters (in snake_case)</param>
+        /// <returns></returns>
+        public static string GetQuery(List<string> columns, List<string> where)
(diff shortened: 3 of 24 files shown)


sqe-http-api/Controllers/ArtefactController.cs
```diff
@@ -99,10 +99,10 @@ public async Task<ActionResult<ArtefactDTO>> GetArtefact([FromRoute] uint editio
         }
         
         /// <summary>
-        /// Provides a listing of all artefacts that are part of the specified edition, including data about
-        /// the related images
+        /// Creates a new artefact with the provided data.
```
Bronson-Brown-deVost, line 111: I guess I fixed some documentation here.

sqe-http-api/Controllers/ArtefactController.cs
```diff
@@ -124,19 +124,5 @@ public async Task<ActionResult<ArtefactDTO>> GetArtefact([FromRoute] uint editio
                 return NotFound();
             }
         }
-
```
Bronson-Brown-deVost, line 127: I removed these, which are no longer used.

sqe-http-api/DTOs/Artefact.cs
```diff
@@ -38,4 +38,19 @@ public class ArtefactDesignationListDTO
     {
         public List<ArtefactDesignationDTO> artefactDesignations { get; set; }
     }
+
```
Bronson-Brown-deVost, line 41: I put these DTOs in the proper place.

data-access/data-access.csproj
```diff
@@ -14,4 +14,10 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Abstractions" Version="2.2.0" />
     <PackageReference Include="Wkx" Version="0.5.0" /> 
   </ItemGroup>
+
```
Bronson-Brown-deVost: I am not really sure why this is here.

data-access/data-access.csproj
```diff
@@ -14,4 +14,10 @@
...
+  <ItemGroup>
+    <Reference Include="Microsoft.Extensions.Options, Version=2.2.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60">
+      <HintPath>..\..\.nuget\packages\microsoft.extensions.options\2.2.0\lib\netstandard2.0\Microsoft.Extensions.Options.dll</HintPath>
+    </Reference>
+  </ItemGroup>
```
Bronson-Brown-deVost: I am not really sure why this is here.

Bronson-Brown-deVost: It is unnecessary, I have removed it.

data-access/Queries/User.cs
```diff
@@ -37,4 +52,139 @@ public class Return
...
+    /// Checks whether an account with @Username or @Email exists and whether it is authenticated yet.
+    /// </summary>
+    internal static class CheckUserAuthentication
+    {
+        public const string GetQuery = @"
+SELECT user_id, authenticated, user_name, email
+FROM user
+WHERE user_name = @Username OR email = @Email";
+
+        public class Return
```
zmbq: Don't we usually call this "Result"?

data-access/Queries/User.cs
```diff
@@ -37,4 +52,139 @@ public class Return
...
+        {
+            public int user_id { get; set; } 
+            public bool authenticated { get; set; }
```
zmbq: "authenticated" is the wrong term. An Authenticated user means a user that has logged in. This has confused me throughout the code. We should "activated" - users activate their accounts.

data-access/Queries/User.cs
```diff
@@ -37,4 +52,139 @@ public class Return
...
+    /// <summary>
+    /// Creates an entry in the user_email_token table for @UserId with the token @Token for the request type @Type
+    /// (CreateUserEmailTokenQuery.Activate or CreateUserEmailTokenQuery.ResetPassword).
+    /// </summary>
+    internal static class CreateUserEmailTokenQuery
+    {
+        public const string GetQuery = @"
+INSERT INTO user_email_token (user_id, token, type)
+VALUES(@UserId, @Token, @Type)";
+
```
zmbq, line 119: We should add an expiration date for tokens. A token can't be used if it has expired.

data-access/UserRepository.cs
```diff
@@ -11,14 +15,33 @@ public interface IUserRepository
...
+        /// <summary>
+        /// Returns user information based on the submitted credentials.
+        /// </summary>
+        /// <param name="userName">The user's username</param>
+        /// <param name="password">The user's password</param>
+        /// <returns></returns>
         public async Task<User> GetUserByPasswordAsync(string userName, string password)
         {
-            var sql = @"SELECT user_name, user_id FROM user WHERE user_name =@UserName AND pw = SHA2(@Password, 224)";
+            var sql = @"SELECT user_name, user_id FROM user WHERE user_name = @UserName AND pw = SHA2(@Password, 224) AND authenticated = 1";
```
zmbq: This probably means users who haven't activated their accounts cannot log in. This is not ideal. Users with inactive accounts *should* be able to log in with their password, and then see a message asking them to activate their account. They will also have a link for resending the activation email.

data-access/UserRepository.cs
```diff
@@ -44,5 +73,262 @@ public async Task<UserEditionPermissions> GetUserEditionPermissionsAsync(uint us
...
+        }
+        
+        /// <summary>
+        /// Creates a request in the database for a reset password token.
+        /// </summary>
+        /// <param name="email">Email address of the user who has forgotten the password</param>
+        /// <returns>Returns user information and secret token that are used to format the reset password email</returns>
+        public async Task<User> RequestResetForgottenPasswordAsync(string email)
+        {
+            await CleanOldTokens(); // Always flush the expired tokens from the database first
```
zmbq: This function should be called `CleanOldTokensAsync`.

data-access/UserRepository.cs
```diff
@@ -44,5 +73,262 @@ public async Task<UserEditionPermissions> GetUserEditionPermissionsAsync(uint us
...
+
+        /// <summary>
+        /// Clears out old tokens, but see the comments to DeleteOldTokens.
+        /// </summary>
+        /// <returns></returns>
+        private async Task CleanOldTokens()
+        {
+            using (var connection = OpenConnection())
+            {
+                await connection.ExecuteAsync(DeleteOldTokens.GetQuery, new { Days = _tokenValidDays });
```
zmbq: I see now that tokens do have expiry dates, but I couldn't see them in the queries.

sqe-http-api/Controllers/UserController.cs
```diff
@@ -20,34 +23,126 @@ public UserController(IUserService userServiceAuthenticate)
...
+            {
+                return Conflict(new {message = err.Message}); // I would rather not tell which element is wrong (again someone might phish for valid emails), but maybe it is not a problem.
+            }
+        }
+        
+        /// <summary>
+        /// Confirms creation of new user account.
+        /// </summary>
+        /// <param name="payload">JSON object with token from user registration email.</param>
+        /// <returns>Returns 200 for success and 404 for failure.</returns>
```
zmbq: If the result is empty, we should return a 204.

sqe-http-api/Controllers/UserController.cs
```diff
@@ -20,34 +23,126 @@ public UserController(IUserService userServiceAuthenticate)
...
         /// <param Name="userParam">JSON object with a username and password parameter</param>
         [AllowAnonymous]
-        [HttpPost("login")] // api/v1/user/login
+        [HttpPost("login")]
         public async Task<ActionResult<LoginResponseDTO>> AuthenticateAsync([FromBody]LoginRequestDTO userParam)
         {
             var user = await _userService.AuthenticateAsync(userParam.userName, userParam.password);
             if (user == null)
-                return Unauthorized(new { message = "Username or password is incorrect", code = 600}); // TODO: Add Error Code
+                return Unauthorized(new { message = "Username or password is incorrect ", code = 600}); // TODO: Add Error Code
```
zmbq: You added an unnecessary space. Also, the error code has already been added, so you can delete the TODO

This serves as the petty review comment of the day

sqe-http-api/Controllers/UserController.cs
```diff
@@ -20,34 +23,126 @@ public UserController(IUserService userServiceAuthenticate)
...
-            return user;
+            await _userService.RequestResetLostPasswordAsync(payload.email);
+            return NoContent();
+        }
+        
+        /// <summary>
+        /// Change the password for the currently logged in user.
+        /// </summary>
+        /// <param name="payload">A JSON object with the old password and the new password.</param>
+        /// <returns>Status 200 with a successful request or status 409 for an unsuccessful request</returns>
```
zmbq: Again, return 204 for empty successful queries

sqe-http-api/DTOs/User.cs
```diff
@@ -22,4 +22,54 @@ public class LoginResponseDTO : UserDTO
     {
         public string token { get; set; }
     }
+    
+    public class NewUserDTO
```
zmbq: Why is this different from the regular UserDTO?

sqe-http-api/DTOs/User.cs
```diff
@@ -22,4 +22,54 @@ public class LoginResponseDTO : UserDTO
...
+        /// <param name="email">Email address for the new user account, must be unique</param>
+        /// <param name="password">Password for the new user account</param>
+        /// <param name="organization">Name of affiliated organization (if any)</param>
+        /// <param name="forename">The user's given name (may be empty)</param>
+        /// <param name="surname">The user's family name (may be empty)</param>
+        public NewUserDTO(string userName, string email, string password, string organization, string forename, string surname)
+        {
+            this.userName = userName;
+            this.email = email;
+            this.password = password;
```
zmbq: Don't ever ever return passwords from the backend back to the frontend.

sqe-http-api/Helpers/SystemEmailer.cs
```diff
@@ -0,0 +1,55 @@
...
+    public class DevEmailSender : IEmailSender
+    {
+        /// <summary>
+        /// Sends an email
+        /// </summary>
+        /// <param name="email">Email address to send the message to.</param>
+        /// <param name="subject"></param>
+        /// <param name="htmlMessage"></param>
+        /// <returns></returns>
+        public async Task SendEmailAsync(string email, string subject, string htmlMessage)
```
zmbq, line 22: Do we want some code that takes Razor templates and turns them into emails? Or is this overkill?

sqe-http-api/Services/UserService.cs
```diff
@@ -19,28 +19,40 @@ public interface IUserService
         UserDTO GetCurrentUser();
         uint? GetCurrentUserId();
         UserInfo GetCurrentUserObject(uint? editionId = null);
+        Task<UserDTO> CreateNewUserAsync(NewUserDTO newUserData);
```
zmbq: Ooooh. It's the `NewUserRequestDTO` ! You should rename the class, then.

sqe-http-api/Services/UserService.cs
```diff
@@ -104,16 +122,110 @@ public UserDTO GetCurrentUser()
...
+        /// <returns>Returns a UserDTO for the newly created user account.</returns>
+        public async Task<UserDTO> CreateNewUserAsync(NewUserDTO newUserData)
+        {
+            // Ask the repo to create the new user
+            var createdUser = await _userRepository.CreateNewUserAsync(newUserData.userName, newUserData.email, newUserData.password,
+                newUserData.forename, newUserData.surname, newUserData.organization);
+            
+            // Email the user
+            // TODO: Add link to web endpoint when we know what that is. Can token be in URL query?
+            const string emailBody = @"
```
zmbq, line 216: This should probably go into another file, a resource maybe.

sqe-http-api/Services/UserService.cs
```diff
@@ -104,16 +122,110 @@ public UserDTO GetCurrentUser()
...
+<html><body>Dear $User,<br>
+<br>
+Thank you for registering with the Scripta Qumranica Electronica research platform.  The token
+to activate your new account is: $Token.<br>
```
zmbq, line 220: Also, don't place tokens in the email, you should put a link that user clicks (I understand you used this for debugging purposes)

sqe-http-api/Services/UserService.cs
```diff
@@ -104,16 +122,110 @@ public UserDTO GetCurrentUser()
...
+            var userInfo = await _userRepository.RequestResetForgottenPasswordAsync(email);
+            if (userInfo == null) // Silently return on error
+                return;
+            
+            // Email the user
+            // TODO: Add link to web endpoint when we know what that is. Can token be in URL query?
+            const string emailBody = @"
+<html><body>Dear $User,<br>
+<br>
+Sorry to hear that you have lost your password for Scripta Qumranica Electronica.  You may reset your password with the token: $Token.<br>
```
zmbq, line 280: Same here. A link instead of a token.

sqe-http-api/Services/UserService.cs
```diff
@@ -104,16 +122,110 @@ public UserDTO GetCurrentUser()
...
+        /// <summary>
+        /// Change password from old password to new password for the user's account.
+        /// </summary>
+        /// <param name="user">User object with all information for the user requesting a password change</param>
+        /// <param name="oldPassword">The old password for the user's account</param>
+        /// <param name="newPassword">The new password for the user's account</param>
+        /// <returns></returns>
+        public async Task ChangePasswordAsync(UserInfo user, string oldPassword, string newPassword)
+        {
+            await _userRepository.ChangePasswordAsync(user, oldPassword, newPassword);
```
zmbq, line 302: Do we want to email someone when their password has been changed? A lot of services do that.

zmbq (changes requested):

data-access/UserRepository.cs
Bronson-Brown-deVost: Good point, though I disagree about being able to login.  The API should not log in valid but unactivated users (i.e., send them a JWT), it should instead send a response letting them know they are authenticated, but not authorized, and need still need to activate (I guess HTTP error 403 applies here).  We can add an endpoint for a request to resend the activation email (user provides the email address). Let me know what you think, and I can add this.

data-access/Queries/User.cs
Bronson-Brown-deVost: Yep, thanks.

data-access/Queries/User.cs
Bronson-Brown-deVost: You're right.  I was think authenticated in terms of the Email address being authenticated.  I have changed this to activated in the database and in the API (along with renaming of classes/methods).

Bronson-Brown-deVost: Done.

data-access/Queries/User.cs
Bronson-Brown-deVost: Done.

data-access/Queries/User.cs
Bronson-Brown-deVost: See below. 
Done.

data-access/UserRepository.cs
Bronson-Brown-deVost: I've removed all the "clean old token" code, it does not belong in the API.  The database itself deletes tokens once they have expired (I have set the expiration time to two days).

sqe-http-api/Controllers/UserController.cs
Bronson-Brown-deVost: Thanks, I have updated the code and the description in our Website API Google Doc.

sqe-http-api/Controllers/UserController.cs
Bronson-Brown-deVost: Thanks.

Done.

sqe-http-api/Controllers/UserController.cs
Bronson-Brown-deVost: Thanks, I have updated the code and the description in our Website API Google Doc.

sqe-http-api/DTOs/User.cs
Bronson-Brown-deVost: Renamed according to comment below.

Done.

sqe-http-api/DTOs/User.cs
Bronson-Brown-deVost: This was a request DTO, not a response DTO.  The DTO has been renamed according to the comment below.

Done.

sqe-http-api/Services/UserService.cs
Bronson-Brown-deVost: Thanks.  It is renamed now.

Done.

sqe-http-api/Services/UserService.cs
Bronson-Brown-deVost: Agreed (see the TODO in line 137).

sqe-http-api/Services/UserService.cs
Bronson-Brown-deVost: Agreed (~line 177).

sqe-http-api/Services/UserService.cs
Bronson-Brown-deVost: That is a nice idea.  I have added that.

Done.

sqe-http-api/Helpers/SystemEmailer.cs
Bronson-Brown-deVost: Its not a bad idea, especially if we will be sending several types of emails, then we could use a standardized template.  I would like to do this, but I will push it off for now.

sqe-http-api/Services/UserService.cs
Bronson-Brown-deVost: Yeah, I like the idea you mentioned above to use a Razor template, so I am going to push this off for now till we have that set up.

data-access/UserRepository.cs
zmbq: The solution of returning 200 with no token is good enough.

data-access/Models/ScrollModels.cs
```diff
@@ -12,7 +12,7 @@ public class Edition
         public bool Locked { get; set; }
         public bool IsPublic {get; set;}
         public DateTime? LastEdit { get; set; }
-        public User Owner { get; set; }
+        public UserToken Owner { get; set; }
```
zmbq, line 15: Why are we returning the owner's token? I see no good use cases for this, just bad ones.

data-access/Models/ScrollModels.cs
```diff
@@ -32,7 +32,7 @@ public class Permission
 
     public class Share
     {
-        public User User { get; set; }
+        public UserToken UserToken { get; set; }
```
zmbq, line 35: Same question here - why return a Token?

data-access/Models/UserModels.cs
```diff
@@ -7,8 +7,33 @@ public class User
...
+        public string Organization { get; set; }
+        public string Email { get; set; }
+    }
+
+    public class UserEmail : User
+    {
+        public string Forename { get; set; }
+        public string Surname { get; set; }
+        public string Email { get; set; }
+    }
```
zmbq, line 36: I'm beginning to think there are too many user DTOs here, making it cumbersome. Let's not change it for now, but let's not add any new DTOs but rather organize these (or simplify them).

data-access/Queries/User.cs
```diff
@@ -1,31 +1,64 @@
...
+WHERE $Where";
 
-        public User CreateModel()
+        /// <summary>
+        /// Returns a formatted query on the user table
+        /// </summary>
+        /// <param name="columns">Names of the columns to be retrieved (in snake_case)</param>
+        /// <param name="where">Names of the where parameters (in snake_case)</param>
+        /// <returns></returns>
+        public static string GetQuery(List<string> columns, List<string> where)
```
zmbq: Now I understand the ToPascalCase. I think this is really overkill. Simply return all the columns and pick the ones you decide to copy to the DTO. We will never face performance issues due to queries of the user table.

data-access/UserRepository.cs
```diff
@@ -39,19 +47,45 @@ public UserRepository(IConfiguration config) : base(config)
...
-                var results = await connection.QueryAsync<UserQueryResponse>(sql, new
+                return await connection.QuerySingleAsync<UserLoginResponse>(LoginQuery.GetQuery, new
                 {
                     UserName = userName,
                     Password = password
                 });
+            }
+        }
+
+        public async Task<DetailedUserToken> GetDetailedUserById(UserInfo userInfo)
```
zmbq: OK, now there are too many repository methods, too, and probably controller endpoints. Why don't we have just two user DTOs - one with all the fields and the other with all the fields *and* the token, and we always return all of them? Everything will be simpler - the API documentation, the queries, the repository, the service, the controller.

sqe-http-api/DTOs/User.cs
```diff
@@ -20,10 +20,11 @@ public class UserDTO
 
     public class LoginResponseDTO : UserDTO
     {
-        public string token { get; set; }
+        public string token { get; set; } // Used with successful login to activated account
```
zmbq, line 23: Isn't this really the *only place* where we return a token? And nowhere else? So we don't really need a token in the UserDTO - just all the fields?

zmbq (changes requested):

Bronson-Brown-deVost:
Ok, I think this PR has been worked enough for now.  So let's pull it in.
I have tested quite a few user scenarios, and all seems to be working as expected.  I would like to write tests for these possibilities to make sure (but we will need to review and pull the integration-tests branch).

zmbq (approved):

PR #7, opened 2019-06-19
Integration tests

Bronson-Brown-deVost:


.githooks/pre-commit
```diff
@@ -0,0 +1,50 @@
+#!/bin/sh
```
zmbq, line 1: Will this run properly on Windows?
Do we want these to apply just to specific branches?

api-test/EditionTests.cs
```diff
@@ -0,0 +1,267 @@
...
+using System.Threading.Tasks;
+using api_test.Helpers;
+using Bogus;
+using Dapper;
+using Microsoft.AspNetCore.Mvc.Testing;
+using SQE.SqeHttpApi.Server;
+using SQE.SqeHttpApi.Server.DTOs;
+using Xunit;
+
+namespace api_test
```
zmbq: Should we perhaps use a full namespace? Perhaps `SQE.SqeHttpApi.Tests` ?

sqe-http-api/Controllers/EditionController.cs
```diff
@@ -76,25 +68,11 @@ public async Task<ActionResult<EditionListDTO>> ListEditions()
         [ProducesResponseType(404)]
         public async Task<ActionResult<EditionDTO>> UpdateEdition([FromBody] EditionUpdateRequestDTO request, [FromRoute] uint editionId)
```
zmbq, line 69: This endpoint, as well as others, do not seem to be documented in our API doc. Can you please add all these endpoints and additions to DTOs?

data-access/Helpers/Exceptions.cs
```diff
@@ -1,38 +1,151 @@
...
 {
-    // TODO: Decide whether the controller is aware of these exceptions, or the service translates them
-    public abstract class RepositoryException: Exception
+    // Exceptions of this class will be caught by the middleware and thrown back to HTTP requests with the proper
+    // response codes.  This enables a RESTful experience regardless of whether a request is made via HTTP or by
+    // websocket.  In either case the HTTP router or the Hub will catch a child of the HttpException class
+    // and return that information to the user.
+    // Errors should be thrown properly at the lowest levels of code, this will reduce boiler-plate in higher
+    // level functions like the Services and ensure more consistent responses.
+    public class HttpException : Exception
```
zmbq: Is `HttpException` the right name? It is thrown by repositories, which are also used with SignalR.

data-access/Helpers/Exceptions.cs
```diff
@@ -1,38 +1,151 @@
...
+    public static class StandardErrors
     {
-        public DbDetailedFailedWrite(string msg) : 
-            base($"Failed writing to DB. {msg}")
-        { }
+
+        #region Permissions errors
+        public static HttpException NoPermissions(uint userId, uint editionId)
+        {
+            return new HttpException(HttpStatusCode.Forbidden, 620,
```
zmbq: We probably need a spreadsheet with all the errors and error codes. It'll be easier when we create an i18ned frontend again.

data-access/Helpers/Exceptions.cs
```diff
@@ -1,38 +1,151 @@
...
+            this.message = message;
+        }
+
     }
     
-    public class DbDetailedFailedWrite: RepositoryException
+    // This is a collection of ready-made errors that can be thrown. They include a standard HTTP status error code
+    // and an internal project error code with accompanying message in English.
+    public static class StandardErrors
     {
```
zmbq, line 79: I like the StandardErrors idea. I would probably use specific exception classes for each exception, so I can catch them. So - `HttpException` ( which needs a better name), `PermissionException` which is also abstract, and then `NoPermissionsException`, `NoReadPermissionsException` and so forth. Then `DataException` and all the exceptions that inherit from it.

That way you can catch all `PermissionException` errors in one swoop, and turn all of them in 403 return codes in an HTTP middleware, or turn them into something else with SignalR, and we decouple the logical problems from the HTTP result codes.

data-access/Helpers/Licence.cs
```diff
@@ -1,5 +1,3 @@
-using SQE.SqeHttpApi.DataAccess.Models;
```
zmbq, line 1: Licence is hurting my eyes. I understand it's correct, but still - hurting my eyes. Don't change it, I'll get better glasses instead.

data-access/Queries/Artefact.cs
```diff
@@ -55,16 +60,20 @@ public static string GetQuery(uint? userId, bool mask = false, bool ordered = tr
...
+            SELECT DISTINCT $Table_id
             FROM $Table
             JOIN $Table_owner USING($Table_id)
-            WHERE $Table.artefact_id = @ArtefactId
+            WHERE $Where
                 AND $Table_owner.edition_id = @EditionId
             ";
 
-        public static string GetQuery(string table)
+        private const string _normalArt = "$Table.artefact_id = @ArtefactId";
```
zmbq, line 70: A couple of comments here would be very helpful - explaining what the different where clauses do.

sqe-http-api/Controllers/ArtefactController.cs
```diff
@@ -42,14 +44,7 @@ private void ParseOptionals(List<string> optionals, out bool images, out bool ma
...
             ParseOptionals(optional, out var images, out var masks);
-            try
-            {
-                return await _artefactService.GetEditionArtefactListingsAsync(_userService.GetCurrentUserId(), editionId, masks, images);
-            }
-            catch (NotFoundException)
-            {
-                return NotFound();
-            }
+            return await _artefactService.GetEditionArtefactListingsAsync(_userService.GetCurrentUserId(), editionId, masks, images);
```
zmbq, line 47: Niiice! Much shorter with the new exceptions.

sqe-http-api/Helpers/HttpExceptionMiddleware.cs
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        public async Task Invoke(HttpContext context)
+        {
+            try
+            {
+                await this.next.Invoke(context);
+            }
+            catch (HttpException httpException)
+            {
```
zmbq, line 33: If we through different exceptions, you can catch all permission exceptions and set the code to 403, all data exceptions and set the code to 409 (or whatever makes sense).

sqe-http-api/Helpers/InputValidation.cs
```diff
@@ -0,0 +1,57 @@
...
+                if (mat.Length != 2)
+                    return false;
+                if (mat[0].Length != 3)
+                    return false;
+                if (mat[1].Length != 3)
+                    return false;
+                return true;
+            }
+
+            bool IsValidValues(double[] row)
```
zmbq: A comment here, what consists of valid values, would be helpful

sqe-http-api/Helpers/InputValidation.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                if (row[0] < -1 && row[0] > 1)
+                    return false;
+                if (row[1] < -1 && row[1] > 1)
+                    return false;
+                return (int)row[2] == row[2];
```
zmbq: Are you testing that `row[2]` contains an integer? This may fail due to rounding errors. It will also fail with negative numbers - (int)-0.99999 is 0 and not -1. 

Better check that Math.Abs(Math.round(row[2]) - row[2]) < 1e-6`  (I'm using Round so it works with negative numbers as well)

zmbq: Also, note that I'm asking what it is you're testing - that's because a comment would have helped me.

sqe-http-api/Helpers/SystemEmailer.cs
```diff
@@ -2,56 +2,93 @@
...
+            // Don't bother trying to send an email unless we have all the necessary email settings
+            // TODO: Should we throw an error for this, or even kill the whole program?
+            if (string.IsNullOrEmpty(_config.GetConnectionString("MailerEmailAddress"))
+                || string.IsNullOrEmpty(_config.GetConnectionString("MailerEmailUsername"))
+                || string.IsNullOrEmpty(_config.GetConnectionString("MailerEmailPassword"))
+                || string.IsNullOrEmpty(_config.GetConnectionString("MailerEmailSmtpUrl"))
+                || string.IsNullOrEmpty(_config.GetConnectionString("MailerEmailSmtpPort"))
+                || string.IsNullOrEmpty(_config.GetConnectionString("MailerEmailSmtpSecurity"))
+                )
+                return;
```
zmbq: Didn't we say we don't want such configuration errors to be swallowed silently?

zmbq (changes requested):

.githooks/pre-commit
Bronson-Brown-deVost: Good question, it is pretty basic shell stuff, but you never know. I will try to check in my windows media computer. It's really there for the safety of the developer to make sure they don't accidentally push their password up onto GitHub.

api-test/EditionTests.cs
Bronson-Brown-deVost: Yeah, I think that is a good idea. I will change it.

data-access/Helpers/Exceptions.cs
Bronson-Brown-deVost: I'm open to another name.  You are right, these will get caught by signalr too and sent on to the user via websocket connections too.  I guess I thought of them as HttpExceptions since they map a more extensive internal error code system to the HTTP error codes. Do you prefer a different name?

data-access/Helpers/Exceptions.cs
Bronson-Brown-deVost: Good idea!  Let me start a Googledocs spreadsheet.  Do you have a preference for spacing things out and maybe creating code domains? I left some small holes, but they are not too big.

sqe-http-api/Helpers/InputValidation.cs
Bronson-Brown-deVost: I will add it.

sqe-http-api/Helpers/SystemEmailer.cs
Bronson-Brown-deVost: Yeah, that is what I was wondering about in the TODO question.  I guess, my feeling is that in production I think I want to exit the program if no credentials are available for the SMTP server. In development, I am happy to let the program run without any SMTP credentials at all (since I am often working on other things and don't care about the email system). So I guess, I should completely remove these checks from the emailer and instead put some logic near startup that checks them on startup when running in production mode (and gracefully exits with a clear error message when they are not).

sqe-http-api/Helpers/InputValidation.cs
Bronson-Brown-deVost: The test here is to make sure that the number in row[2] is a whole number.  So if row[2] is -0.999 (which is an invalid value), then (int)row[2] == row[2] -> 0 == -0.999 -> false, which is indeed the behavior I want (some documentation would indeed have helped there, sorry).  Another example where row[2] is -1.000, then (int)row[2] == row[2] -> -1 == -1.000 -> true. Basically, the test ensures that there are no significant decimal values, since JSON only has Number type, not int/float, and I cannot otherwise ensure that this value is the equivalent of an int (even if it has trailing 0's).

data-access/Helpers/Exceptions.cs
zmbq: How about `ServerException` or `APIException`?

data-access/Helpers/Exceptions.cs
zmbq: I prefer having codes as strings, anyway, so you can understand what the error is from the code itself.

sqe-http-api/Helpers/InputValidation.cs
zmbq: JSON has a number type, but C# has ints. If you put an `int[][]` in the DTO, aren't floats going to be rejected by ASP.NET Core?

data-access/Helpers/Exceptions.cs
Bronson-Brown-deVost: I like APIException.  Also, I think you mentioned that each exception should be its own (sub)class, I liked that idea too.  It should not be too hard to implement that.

sqe-http-api/Helpers/SystemEmailer.cs
zmbq: The server can run on development without the email, but you must return an error if someone tries to register a new user on a development machine with no email. They will not be able to do that, and they need to see it. Otherwise we can expect 5 or 6 lost hours figuring out where are all those activation emails.

data-access/Helpers/Exceptions.cs
zmbq: Good. Just group them into groups of exceptions, so all permissions errors derive from `APIPermissionException`. It will allow you catch some of them in your middleware without writing numerous catch clauses (that are going to malfunction if you add another exception of that type). Of course `APIPermissionException` itself is derived from `APIException`

zmbq: I do not have strong feelings towards taking the HTTP code out of these classes and limiting it to the middleware. It's obviously the more organized thing to do, but I think we can safely ignore this one. Putting the HTTP code inside the Exception implementation is probably a bit easier.

.githooks/pre-commit
Bronson-Brown-deVost: I am testing now.  The script runs properly in git bash.  However, you need some SH extension to run it in powershell. I don't know enough about how people use GIT on windows.  Do you always use the GIT gui, or git bash for pushing/pulling? I will need to do a bit more testing and reading before I am confident we have this working properly.

zmbq: Well, it depends. I mostly use git from Powershell, and sometimes from whatever editor I use. In rare occasions I use TortoiseGit as the GUI. Others here never use git from Powershell and always use the GUI. I'm sure there are others that use other GUIs as well. I think we shouldn't require people to install something in addition to git - some of them just have whatever is in Visual Studio and don't even install their own standalone gits.

zmbq: Plus, we have code reviews, if someone puts their password there - we'll catch it and let them know. So don't spend too much time on this - it's not a big problem.

Bronson-Brown-deVost: Ok. To my surprise it just works in powershell too and in the visual studio code built-in git interface.  Anyone who wants to use the hook must run ' git config core.hooksPath .githooks' once in the repo after they clone it.  I will check to see if there is a way to make this happen automatically.

Bronson-Brown-deVost: Thanks for the input.  I think I will leave this for now as you suggest and add a suggestion in the readme that developers might want to run `git config core.hooksPath .githooks` to help ensure that they don't upload sensitive information.  If they do not use the script, or find some other way to accidentally transmit email credentials, then hopefully they or someone else will catch it (we only review merges to master and development, so commits to other branches might not be seen by another human).
FWIW Automatically setting `git config core.hooksPath .githooks` seems to be an anti-pattern—apparently a repo should not be mucking about with local config settings, I can imagine many good reasons why this is the case.

data-access/Helpers/Exceptions.cs
Bronson-Brown-deVost: Ah, so are you suggesting that we remove the internal code numbers altogether and just use a descriptive English string as the code?

zmbq: Yes. NO_PERMISSIONS is a very good code for a NoPermissions exception.

sqe-http-api/Helpers/InputValidation.cs
Bronson-Brown-deVost: I think it isn't a real problem, the transform matrix in JSON is a 2d array of Numbers, the DTO just uses receives that JSON string and inserts it as-is into the database, where this data is also just a string, so the data is never converted into a C# data structure for insertion into the database.  
The point of the validator here is to make sure that any submitted transform matrix string is in fact 1. valid json, 2. having a "matrix" property that is a 2d array of doubles, 3. which has exactly two first level members of the array (i.e., [[],[]] and not [[],[],[]], etc.), 4. each of which has exactly three members, 5. the first two of which fall within a specific a specific range of values (I need to recheck what is a valid range for these), and 6. the third member is a whole number—it doesn't matter if it is 3.000 or 3, I just want it to be a whole number because sub-pixel translation makes no sense here (e.g., translate on the X axis by 5493.8796 pixels).
I will add all this to the function to clarify and also I will add it to the API documentation.

sqe-http-api/Helpers/SystemEmailer.cs
Bronson-Brown-deVost: Ok.  I made the change now.  I put in checks in the `Startup` method in `Startup.cs`, but that feels wrong to me.  Where would you put such checks to verify that the settings in appsettings.json are valid?  I use `if (env.IsProduction())` to only run these tests in production.  Also what would you do to stop the program in the case that one of these checks finds an unset variable, I used `throw new SystemException` which crashes the program as it is an uncaught exception.

On the other hand. I could forget those checks on startup altogether and just let MailKit throw it's error; there is no guarantee after all that the settings in appsettings.json are correct just because they are there.  It is setup now in SystemEmailer.cs that in production it will catch any send errors and throw a nice message back to the user saying they probably input an invalid email (I still need to add a server-side log), otherwise in development it sends the whole error on.

zmbq: That's a good question. Actually, that's a few good questions. First - about the placement of the code, putting it in Startup.cs (or at least - calling it from Startup.cs) seems OK. 

The bigger question is which checks we're going to perform at startup, so that we know the server is not properly configured right away, instead of later on. There are a few things that need to be configured properly: the database, redis for SignalR, the email configuration and perhaps more in the future.

I think it's more important to add logging than to check these things at startup - because these thing can fail after startup, too (the mail server can die, for instance). I guess at startup we just need to check that the configuration is complete, and log a warning if it isn't on dev, or stop the server on prod (and log a critical error).

sqe-http-api/Helpers/InputValidation.cs
zmbq: OK.

sqe-http-api/Helpers/SystemEmailer.cs
Bronson-Brown-deVost: Many thanks.  I will work on implementing that now.

api-test/EditionTests.cs
Bronson-Brown-deVost: Ok. This has been implemented now.

data-access/Helpers/Exceptions.cs
Bronson-Brown-deVost: Ok. They now follow what one would expect from common OOP practice.  I have an abstract base class, then abstract subclasses for the groups of exceptions, and then each individual exception class inherits from one of those.  As for the error codes (as mentioned below), I use the class name as the internal error code, which is simple enough and probably a nice feature for debugging and avoiding naming errors (you can't have two classes with the same name in the same namespace).

sqe-http-api/Helpers/HttpExceptionMiddleware.cs
Bronson-Brown-deVost: Yeah, I wrote a little middleware that catches these exceptions, reads the HTTP status code, and then packages the exception as that type of HTTP response to be sent back to the user.  There will be a similar component in the SignalR code that packages these exceptions in an appropriate way for the realtime users.

sqe-http-api/Helpers/InputValidation.cs
Bronson-Brown-deVost: I cleaned this up and added documentation.  I also moved the TransformMatrix class to the data-access/Models section, since it makes more sense there.

sqe-http-api/Helpers/SystemEmailer.cs
Bronson-Brown-deVost: I now created a class `StartupChecks` which gets run at startup.  It always checks that that database can be accessed, and when running in production mode it tests that it can connect to the SMTP server.  In both cases it not only checks for the necessary settings in appsettings.json, but also makes a real connection to the external resource.  So that should help in killing the program immediately if the settings are invalid or the resource is unreachable.  Any runtime errors are caught in the application code now and will manifest in errors sent to the client, but I have not yet added server side logging for them.  I don't want to add logging to this PR, I think we covered enough ground already.

zmbq (approved):

PR #10, opened 2019-07-22
Text data

Bronson-Brown-deVost:
This includes retrieval of text transcriptions (including finding all text fragments and lines in an edition).  It also includes the create text fragment endpoint and the relevant integration tests. I will continue to add text mutation endpoints described in the web API google doc, but I will do them in separate branches to minimize the complexity of each PR.


api-test/EditionTests.cs
```diff
@@ -256,7 +256,7 @@ FROM edition
             
             var randomEdition = allUserEditions[_faker.Random.Int(0, allUserEditions.Count - 1)];
             var url = $"/{version}/editions/{randomEdition}";
-            var (response, editionResponse) = await HttpRequest.SendAsync<string, EditionListDTO>(_client,
+            var (response, editionResponse) = await HttpRequest.SendAsync<string, EditionGroupDTO>(_client,
```
zmbq, line 259: I actually missed the C# multiple return types feature. It's nice!

zmbq (approved):

PR #12, opened 2019-07-26
Edition permissions

Bronson-Brown-deVost:
Adds endpoints for edition sharing and setting the permissions associated with that.  It also updates the logic for deleting editions and cleans up some of the integration tests.

### DIFF
--- a/api-test/Helpers/EditionHelpers.cs
+++ b/api-test/Helpers/EditionHelpers.cs
@@ -4,6 +4,7 @@ using System.Threading.Tasks;
 using Bogus;
 using Dapper;
 using SQE.SqeHttpApi.Server.DTOs;
+using Xunit;
 
 namespace SQE.ApiTest.Helpers
 {
@@ -18,7 +19,7 @@ namespace SQE.ApiTest.Helpers
         /// </summary>
         /// <param name="userId">Id of the user whose editions should be randomly selected.</param>
         /// <param name="jwt">A JWT can be added the request to access private editions.</param>
-        /// <returns></returns>
+        /// <returns>a randomly selected EditionDTO</returns>
         public static async Task<EditionDTO> GetRandomEdition(DatabaseQuery db, HttpClient client, uint userId = 1, string jwt = null)
         {
             const string sql = @"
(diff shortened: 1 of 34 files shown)


api-test/Helpers/EditionHelpers.cs
```diff
@@ -18,7 +19,7 @@ public static class EditionHelpers
         /// </summary>
         /// <param name="userId">Id of the user whose editions should be randomly selected.</param>
         /// <param name="jwt">A JWT can be added the request to access private editions.</param>
-        /// <returns></returns>
+        /// <returns>a randomly selected EditionDTO</returns>
```
zmbq, line 22: Why do we need a randomly selected edition?

data-access/EditionRepository.cs
```diff
@@ -149,7 +157,10 @@ public async Task ChangeEditionNameAsync(UserInfo user, string name)
             // until the copy process was complete in order to guard against
             // creating an inconsistent copy.
             // What if someone unlocks the source scroll mid-copy?
-            using (var transactionScope = new TransactionScope())
+            using (var transactionScope = new TransactionScope(
+                TransactionScopeOption.Required,
+                new TransactionOptions() { IsolationLevel = System.Transactions.IsolationLevel.ReadUncommitted })
```
zmbq: (I may have left this comment before starting this review, I tried to delete it)
Don't use ReadUncommitted unless you have a *really really* good reason. It's almost always a bad idea. Here you seem to have used it to improve performance - do we have a performance issue when people rename editions? I would drop all ReadUncommitted everywhere. If we are having issues, we'll solve them hopefully without ReadUncommitted.

data-access/EditionRepository.cs
```diff
@@ -236,6 +247,215 @@ public async Task ChangeEditionCopyrightAsync(UserInfo user, string copyrightHol
...
+                    )
+                );
+
+            // This transaction may take a while, so we cannot lock all of these tables. Otherwise, we DO get deadlock.
+            // ReadUncommitted is fine because we will not get any further writes to this edition (all editors have lost
+            // write permission), and any new writes to the user_email_token table will be irrelevant (the token
+            // is unique).
+            using (var transactionScope = new TransactionScope(
+                TransactionScopeOption.Required,
+                new TransactionOptions() { IsolationLevel = System.Transactions.IsolationLevel.ReadUncommitted })
```
zmbq: Here to, no ReadUncommitted please

zmbq (changes requested):
I usually don't have strong feelings about comments. In this case, I really don't want any ReadUncommitted - anywhere. We need a very good reason to use something like this.

api-test/Helpers/EditionHelpers.cs
Bronson-Brown-deVost: I figured that it would keep the integration tests fair by sampling different editions each time.

data-access/EditionRepository.cs
Bronson-Brown-deVost: I am very open to getting rid of ReadUncommitted.  The function here is one that creates a copy of an edition (users will use to branch an edition).  So it often makes many writes to the database, 1000s really, so higher isolation levels that lock the whole table when making so many writes cause real deadlock issues.  I was getting a lot of them in the integration tests.  If I could find some other way to prevent locking the table, that would be a nice solution.  I don't want to rely on catching deadlock errors and automatically retrying the query after a brief pause (though some logic for that should be added), because I see no reason to lock the table when these writes are being performed, the permission system in the API gives all the safeguards we need.

zmbq:
Read Uncommited in this case will mean the copy of the scroll will be inconsistent! It’s a race condition just waiting to happen! Suppose that during this copy, someone performs an edit operation on the scroll – one that needs multiple SQL statements. The copy operation may get a few of these statements but not all of them, giving you a copy that you can’t use. Or worse – an copy with an inconsistency problem you will only see 5 months later when you publish it.

 

  Don’t use Read Uncommitted here.

 

  I’m not sure why you have deadlocks in the tests, perhaps the tests need to be reorganized. The tests do not emulate a real use-case, so this is a test performance problem, not an actual performance problem.

zmbq:
This seems problematic. Suppose a test fails – how do we know which edition caused the problem?

zmbq:
This also means that github may approve one commit for a pull request, and then refuse to merge because the tests succeed once and fail a second time – on the same commit. I think tests should be consistent – if a test fails, it should fail always. Tools assume this is the case.

Bronson-Brown-deVost:
I am now reviewing the documentation on isolation level for MariaDB, it is not the same as PostgreSQL (and other more standard SQL rdbms's) and so falls a bit outside of the conventional wisdom.  I think I want to leave the tests as is, because they represent a high usage situation, which is nice because it provides some minor stress testing.  I agree that unit tests should always produce the same result every time, we can't expect integration tests to work like that unless I have more robust error handling, so I will add more robust error handling.  In any event, I think that I can get things working properly in the API and remove these ReadUncommitted, I really do agree with you that it is a bad idea to use it outside of very specific use cases, and I do not think these usages are such cases.  Thanks for pushing on this.

Bronson-Brown-deVost:
Oh and just to give a clear response to your comment:
> Read Uncommited in this case will mean the copy of the scroll will be inconsistent! It’s a race condition just waiting to happen! Suppose that during this copy, someone performs an edit operation on the scroll – one that needs multiple SQL statements. The copy operation may get a few of these statements but not all of them, giving you a copy that you can’t use. Or worse – an copy with an inconsistency problem you will only see 5 months later when you publish it.

This scenario cannot happen because we only allow copying of locked scrolls and the API does not allow any alterations of locked scrolls (unless Ingo or I try to write directly to the DB without using the API).  So no changes to the scroll could happen during a copy process.  In any event, I will still work to remove the ReadUncommitted, I agree that we should not need to use it here.

Bronson-Brown-deVost:
It appears that my recent database fixes did more to solve the deadlock problem than the ReadUncommitted transactions.  I have removed ReadUncommitted now and the tests continue to pass after multiple tries.  I had thought that the instances of ReadUncommitted must remain in the DatabaseWriter service.  The decision to use it there was the result of a long conversation with Ingo, but it stemmed from your reservations on relying on uniqueness constraints in the database (see the comment in MutateQueries.cs for more information).  In this case we will rely on the database uniqueness constraints to enforce consistency and can now remove the ReadUncommitted isolation level.  Let me know if this will pass review now.  Thanks.

I will add Polly now in a separate branch to catch transient errors from the database and carry out a retry policy before letting the API immediately give up and send the error to the API consumer.  This should help ensure the API is more robust in the face of deadlocks and other brief database outage errors, and any transient errors should be very unlikely to cause an integration test to fail.

zmbq:
I will try to find time and look at it in the next couple of days. It’s 11PM and only now I have time to catch up on what’s been happening at work… It’s good you removed all ReadUncommitted, it shouldn’t be used – ever.

zmbq (approved):

PR #13, opened 2019-07-30
Transient error handling

Bronson-Brown-deVost:
This follows on the heels of the PR for edition-permissions.  It uses Polly to implement a robust retry system for those transient errors where the MariaDB specifications call for a retry (it does not retry any other errors).  It also wraps several transactions in retry policies.  Should all retries fail, the exception then continues to bubble up.  

It also adds a circuit breaker for errors in retrieving a connection from MariaDB when the error reported suggests waiting for a connection to become free before trying again.  In such cases it retries 5 times then pauses all attempts for 5 seconds with the hopes that the server will be less burdened by then.

A follow-up TODO is to add the logging system into this so that we log the retry attempts.  I will add that when I add the logging system.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,5 +15,5 @@ install:
 
 script:
 - dotnet build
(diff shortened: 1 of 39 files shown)


zmbq (commented):
This pull request seems to include changes from the previous edition-permissions pull request. So I can't approve it before I approve the previous one. I hope this doesn't cause a great big mess.

Bronson-Brown-deVost:
Please finish the review for edition-permissions first.  Then we can work through this one.  Thanks.

zmbq:
Please open your Slack.

zmbq (approved):

PR #16, opened 2019-08-13
Updated dtos

Bronson-Brown-deVost:
I have made all the planned updates to include editor id's for each data item.  This uses the latest database version, which appears (in my limited testing) to be able to use a mounted volume to persist data across restarts (and not try to set up the root and user accounts a second time).  I also did some work to further alleviate the deadlock issues. They are much better, but still sometimes present, we will have to discuss other solutions than what has already been tried.

The Travis CI build keeps failing due to issues transmitting the code coverage reports to Coveralls (they are doing a database rebuild or something).  I won't waste any more time on that now, but will retry tomorrow.

Let me know when things look good here and we can merge the PR, then we will introduce the typescript DTO build process into the development toolchain in the typescript-dto branch.


data-access/EditionRepository.cs
```diff
@@ -269,42 +276,55 @@ public async Task<string> DeleteAllEditionDataAsync(UserInfo user, string token)
...
                 editors.Select(
                     x => ChangeEditionEditorRights(user, x.Email, x.MayRead, false, x.MayLock, x.IsAdmin)
-                    )
-                );
-
-            using (var transactionScope = new TransactionScope())
+                )
+            );
+
+            // Note: I had wrapped the following in a transaction, but this has the problem that it can lockup every
```
zmbq, line 282: Why would deleting rows from a table lock the entire table? Is this another MySQL/MariaDB feature?

sqe-dtos/Polygon.cs
```diff
@@ -3,6 +3,8 @@
     public class PolygonDTO
     {
         public string mask { get; set; } 
+        public uint maskEditorId { get; set; }
```
zmbq, line 6: What's the point of having two editor IDs on the same entity?

zmbq (approved):

PR #17, opened 2019-08-18
Delay start

zmbq:
The frontend's docker-compose file fails, because the .NET container is up before the database is. I've added 5 retries (with a 3 second delay between them) before the backend gives up.


Bronson-Brown-deVost (approved):
Glad to see you caught this.  I will probably update it later to use Polly (our standard retry library), but this is fine for now.  Many thanks.

PR #18, opened 2019-08-21
Typescript dto

Bronson-Brown-deVost:
I think we are done here.

I added to your code a git pre-commit script that auto generates the DTO's:
`dotnet build Utilities/GenerateTypescriptDTOs
    dotnet build sqe-dtos
    dotnet Utilities/GenerateTypescriptDTOs/bin/Debug/netcoreapp2.2/GenerateTypescriptDTOs.dll sqe-dtos/bin/Debug/netcoreapp2.2/sqe-dtos.dll ts-dtos/sqe-dtos.ts`

The typescript DTO definitions are found in `ts-dtos/sqe-dtos.ts` and the Readme.md has been updated to point there.  Let me know if there is anything unfinished that you still want to do here before merging.


data-access/Models/TextFragment.cs
```diff
@@ -14,6 +14,7 @@ public class TextFragmentData
     {
         public string TextFragmentName { get; set; }
         public uint TextFragmentId { get; set; }
+        public uint EditionEditor { get; set; }
```
zmbq, line 17: Shouldn't this be EditionEditorId?

zmbq (approved):

data-access/Models/TextFragment.cs
Bronson-Brown-deVost: Yes, it should be.  I'll fix that in the next pull.  Nice catch.

PR #23, opened 2019-09-17
Rois - WIP

Bronson-Brown-deVost:
@zmbq, this branch provides the ROI endpoints you need for v0.4.0.  It also contains the update to the artefact DTO's so they have a `scale`, `rotate`, `translateX`, and `translateY` instead of a transform matrix, as well as the updates to enable setting the log level via docker environment variables.  

I labelled it a WIP because I did not write the integration tests yet.  I though that @daniellaillouz  might want to try doing that (we had talked earlier about maybe having you guys write the API tests).  Then when everything is in order we can pull it in.

There are 7 new endpoints (you can easily explore them with http://localhost:5000/swagger):
ArtefactController.cs
GET /v1/editions/{editionId}/artefacts/{artefactId}/rois (returns all ROIs for an artefact)

RoiController.cs
GET /v1/editions/{editionId}/rois/{roiId} (returns info for the specified ROI)
PUT /v1/editions/{editionId}/rois/{roiId} (updates the specified ROI)
DELETE /v1/editions/{editionId}/rois/{roiId} (deletes the specified ROI)
POST /v1/editions/{editionId}/rois (created a new ROI)
PUT /v1/editions/{editionId}/rois/batch (updates a list of ROIs)
POST /v1/editions/{editionId}/rois/batch (creates a list of ROIs)

Remember that you already get rois for transcribed text with the GET /v1/editions/{editionId}/text-fragments/{textFragmentId} endpoint; we needs some integration tests for this functionality as well.

All the integration testing code is in the api-test project.  Basically if you make your new test class a subclass of `WebControllerTest`, you will get a special HttpClient `_client`, which you can use for all HTTP requests to the API.  There are some other Helpers for your convenience in the Helpers folder (especially SendAsync in HttpRequest.cs).

### DIFF
--- a/data-access/Helpers/Exceptions.cs
+++ b/data-access/Helpers/Exceptions.cs
@@ -229,7 +229,7 @@ namespace SQE.SqeHttpApi.DataAccess.Helpers
 			public DataNotFound(string datatype = null, string id = "0", string searchEntity = null)
 			{
 				if (!string.IsNullOrEmpty(datatype)
-				    && string.IsNullOrEmpty(searchEntity))
+					&& string.IsNullOrEmpty(searchEntity))
 					searchEntity = datatype;
 
 				var fullMsg = string.IsNullOrEmpty(datatype) || id == "0"
--- a/data-access/Helpers/Exceptions.cs
+++ b/data-access/Helpers/Exceptions.cs
@@ -99,16 +99,16 @@ namespace SQE.SqeHttpApi.DataAccess.Helpers
 
 	// This is a collection of ready-made errors that can be thrown. They include a standard HTTP status error code
 	// and an internal project error code with accompanying message in English.
-	public static class StandardErrors
+	public static class StandardExceptions
 	{
 		#region System errors
 
-		public class EmailNotSent : ServerErrorException
+		public class EmailNotSentException : ServerErrorException
 		{
 			private const string customMsg =
 				"Failed sending email to $Email. This is probably a server error and should be reported to the webmaster.";
 
-			public EmailNotSent(string email)
+			public EmailNotSentException(string email)
 			{
 				Error = customMsg
 					.Replace("$Email", email);
@@ -119,11 +119,11 @@ namespace SQE.SqeHttpApi.DataAccess.Helpers
 
 		#region Permissions errors
 
-		public class NoPermissions : ForbiddenDataAccessException
+		public class NoPermissionsException : ForbiddenDataAccessException
 		{
 			private const string customMsg = "User $UserId has no permissions associated with edition $EditionId.";
 
-			public NoPermissions(EditionUserInfo editionUser)
+			public NoPermissionsException(EditionUserInfo editionUser)
 			{
 				Error = customMsg
 					.Replace("$UserId", editionUser.userId.ToString())
@@ -131,11 +131,11 @@ namespace SQE.SqeHttpApi.DataAccess.Helpers
 			}
 		}
 
-		public class NoReadPermissions : ForbiddenDataAccessException
+		public class NoReadPermissionsException : ForbiddenDataAccessException
 		{
 			private const string customMsg = "User $UserId does not have read access to edition $EditionId.";
 
-			public NoReadPermissions(EditionUserInfo editionUser)
+			public NoReadPermissionsException(EditionUserInfo editionUser)
 			{
 				Error = customMsg
 					.Replace("$UserId", editionUser.userId.ToString())
@@ -143,11 +143,11 @@ namespace SQE.SqeHttpApi.DataAccess.Helpers
 			}
 		}
 
-		public class NoWritePermissions : ForbiddenDataAccessException
+		public class NoWritePermissionsException : ForbiddenDataAccessException
 		{
 			private const string customMsg = "User $UserId does not have write access to edition $EditionId.";
 
-			public NoWritePermissions(EditionUserInfo editionUser)
+			public NoWritePermissionsException(EditionUserInfo editionUser)
 			{
 				Error = customMsg
 					.Replace("$UserId", editionUser.userId.ToString())
@@ -155,11 +155,11 @@ namespace SQE.SqeHttpApi.DataAccess.Helpers
 			}
 		}
 
-		public class NoLockPermissions : ForbiddenDataAccessException
+		public class NoLockPermissionsException : ForbiddenDataAccessException
 		{
 			private const string customMsg = "User $UserId is not allowed to lock edition $EditionId.";
 
-			public NoLockPermissions(EditionUserInfo editionUser)
+			public NoLockPermissionsException(EditionUserInfo editionUser)
 			{
 				Error = customMsg
 					.Replace("$UserId", editionUser.userId.ToString())
@@ -167,11 +167,11 @@ namespace SQE.SqeHttpApi.DataAccess.Helpers
 			}
 		}
 
-		public class NoAdminPermissions : ForbiddenDataAccessException
+		public class NoAdminPermissionsException : ForbiddenDataAccessException
 		{
 			private const string customMsg = "User $UserId does not have admin privilege for edition $EditionId.";
 
-			public NoAdminPermissions(EditionUserInfo editionUser)
+			public NoAdminPermissionsException(EditionUserInfo editionUser)
 			{
 				Error = customMsg
 					.Replace("$UserId", editionUser.userId.ToString())
@@ -179,12 +179,12 @@ namespace SQE.SqeHttpApi.DataAccess.Helpers
 			}
 		}
 
-		public class LockedData : LockedDataException
+		public class LockedDataException : Helpers.LockedDataException
 		{
 			private const string customMsg =
 				"Edition $EditionId is locked. User $UserId $Permission admin privilege to unlock it.";
 
-			public LockedData(EditionUserInfo editionUser)
+			public LockedDataException(EditionUserInfo editionUser)
 			{
 				Error = customMsg
 					.Replace("$UserId", editionUser.userId.ToString())
@@ -193,11 +193,11 @@ namespace SQE.SqeHttpApi.DataAccess.Helpers
 			}
 		}
 
-		public class BadLogin : UnauthorizedException
+		public class BadLoginException : UnauthorizedException
 		{
 			private const string customMsg = "Failed login for $Email.";
 
-			public BadLogin(string email)
+			public BadLoginException(string email)
 			{
 				Error = customMsg.Replace("$Email", email);
 			}
@@ -205,11 +205,11 @@ namespace SQE.SqeHttpApi.DataAccess.Helpers
 
 		// Do not use this for login related errors, it is only for actions that require an authenticated user
 		// to resubmit their password.
-		public class WrongPassword : UnauthorizedException
+		public class WrongPasswordException : UnauthorizedException
 		{
 			private const string customMsg = "The password is incorrect.";
 
-			public WrongPassword()
+			public WrongPasswordException()
 			{
 				Error = customMsg;
 			}
@@ -219,14 +219,14 @@ namespace SQE.SqeHttpApi.DataAccess.Helpers
 
 		#region Data errors
 
-		public class DataNotFound : DataNotFoundException
+		public class DataNotFoundException : Helpers.DataNotFoundException
 		{
 			private const string customMsg = "Data not found. ";
 
 			private const string reason =
 				"No entries could be found for $DataType, when searching on $SearchEntity with id $Id.";
 
-			public DataNotFound(string datatype = null, string id = "0", string searchEntity = null)
+			public DataNotFoundException(string datatype = null, string id = "0", string searchEntity = null)
 			{
 				if (!string.IsNullOrEmpty(datatype)
 					&& string.IsNullOrEmpty(searchEntity))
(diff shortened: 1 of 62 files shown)


data-access/Helpers/Exceptions.cs
```diff
@@ -123,59 +123,59 @@ public class NoPermissions : ForbiddenDataAccessException
 		{
 			private const string customMsg = "User $UserId has no permissions associated with edition $EditionId.";
 
-			public NoPermissions(UserInfo user)
+			public NoPermissions(EditionUserInfo editionUser)
```
zmbq: I've obviously missed that in a previous pull request - exception classes usually end with Exception, so `NoPermissionsException`, `NoReadPermissionException`, etc..

data-access/ImageRepository.cs
```diff
@@ -30,8 +30,8 @@ public async Task<IEnumerable<Image>> GetImagesAsync(uint? userId, uint editionI
 					sql,
 					new
 					{
-						UserId = userId ?? 0, // @UserId is not expanded if userId is null
-						EditionId = editionId,
+						UserId = editionUser.userId ?? 0, // @UserId is not expanded if userId is null
```
zmbq: But UserId is never null, it can be 0, but not null. Perhaps it should be nullable? I'm not sure how the code that actually performs the query will look like.

data-access/Models/ArtefactModels.cs
```diff
@@ -7,12 +7,16 @@ public class ArtefactModel
 		public string Name { get; set; }
 		public string Mask { get; set; }
 		public uint MaskEditorId { get; set; }
-		public string TransformMatrix { get; set; }
-		public uint TransformMatrixEditorId { get; set; }
+		public float? Scale { get; set; }
+		public float? Rotate { get; set; }
+		public uint? TranslateX { get; set; }
```
zmbq, line 12: Are the translations always non-negative? Is there no way we'll want to move an artefact that is on the top left corner of the scroll *a little bit* up or left?

data-access/Models/ArtefactModels.cs
```diff
@@ -7,12 +7,16 @@ public class ArtefactModel
...
+		public uint? TranslateY { get; set; }
```
zmbq, line 13: Also, obviously in the database the X and Y values are stored in separate fields. In the model, however, perhaps we can group them in an Offset structure (that has X and Y properties)? For instance, there's no case where TranslateX is null and TranslateY has a value - they're both null or both not null.

data-access/Models/UserModels.cs
```diff
@@ -36,90 +36,47 @@ public class UserEditionPermissions
 		public bool Locked { get; set; }
 	}
 
-	public class UserInfo
+	public class EditionUserInfo
```
zmbq, line 39: I don't like this name. It's too literal. The name should explain the purpose of this class, not what it holds. Unfortunately, I don't have a better name, so let's just keep this one.

data-access/RoiRepository.cs
```diff
@@ -0,0 +1,296 @@
...
+		/// <param name="editionUser">UserInfo object with user details and edition permissions</param>
+		/// <param name="newRois">List of rois to be added to the system.</param>
+		/// <returns></returns>
+		public async Task<List<SignInterpretationROI>> CreateRoisAsync(EditionUserInfo editionUser,
+			List<SetSignInterpretationROI> newRois)
+		{
+			return (await Task.WhenAll(newRois.Select(
+				async (x) =>
+				{
+					var roiShapeId = CreateRoiShapeAsync(x.Shape);
```
zmbq: `roiShapeId` is not really the shape id, it's a task that will eventually resolve the shape id. I think the name should be `roiShapeIdTask`. I was going to call this a bug, but then I saw the `await` a few lines down. So it's not a bug, just a variable with a confusing name.

data-access/RoiRepository.cs
```diff
@@ -0,0 +1,296 @@
...
+			List<SignInterpretationROI> updateRois)
+		{
+			return (await Task.WhenAll(updateRois.Select(
+				async (x) =>
+				{
+					var originalSignRoiInterpretation =
+						await GetSignInterpretationRoiByIdAsync(editionUser, x.SignInterpretationRoiId);
+
+					// TODO: Maybe parse this better, because the strings can be non-equal, but the data may still be the same.
+					var roiShapeId = originalSignRoiInterpretation.Shape == x.Shape
```
zmbq: This, on the other hand, really is a shape id.

data-access/UserRepository.cs
```diff
@@ -184,15 +184,15 @@ public async Task<UserEditionPermissions> GetUserEditionPermissionsAsync(UserInf
 						UserPermissionQuery.GetQuery,
 						new
 						{
-							EditionId = user.editionId,
-							UserId = user.userId
+							editionUser.EditionId,
+							UserId = editionUser.userId ?? 1 // if userID is null we get the permissions for userId 1
```
zmbq: Why? Why do we get the permissions for user 1? This needs a better comment.

docker/deploy/docker-compose.yml
```diff
@@ -37,24 +37,31 @@ services:
             - MYSQL_PASSWORD=mysecretpw
             - MYSQL_DATABASE=SQE
             
-            ## Email settings (account that mails user registration information)
+            ## Email settings (account that mails user registration information), these are required.
```
zmbq, line 40: You should probably take the docker-compose.yml file from the ScrollEditor repository, which takes MAILER_* variables from the environment.

docker/deploy/docker-compose.yml
```diff
@@ -37,24 +37,31 @@ services:
...
             #- WEBSITE_HOST=
               
             ## Dotnet telemetry optout
             - DOTNET_CLI_TELEMETRY_OPTOUT=1
               
             ## Secret string for API JWT generation (the API will error if this is too short)
             - SQE_API_SECRET=kjdsavc987342r8gnx%%6bq1@@lknsdciqeurhfwiurbcea-+++
+            
+            ## Logging levels, valid values are: Verbose, Debug, Information, Warning, Error, Fatal
+            - API_LOGLEVEL=Information # Log level for API generated messages
```
zmbq, line 58: Cool!

sqe-dtos/Artefact.cs
```diff
@@ -26,18 +27,20 @@ public class ArtefactListDTO
...
 
+	// TODO: Should we make the mask and positioning data a PolygonDTO here, like it is in the GET DTO?
 	public class UpdateArtefactDTO
 	{
 		public string mask { get; set; }
 		public string name { get; set; }
-		public string position { get; set; }
+		public float? scale { get; set; }
+		public float? rotate { get; set; }
+		public uint? translateX { get; set; }
```
zmbq: These *definitely* need an OffsetDTO (or TranslationDTO) class. Have two properties in the model is an internal implementation detail, but this is the external API. Grouping the two properties makes sense - they always come together. Plus, as I've written in the comment for the model - they must be both null or both not null, you can't have X null and Y not null. Also, are we sure they can't be negative?

sqe-dtos/Polygon.cs
```diff
@@ -4,7 +4,10 @@ public class PolygonDTO
 	{
 		public string mask { get; set; }
 		public uint maskEditorId { get; set; }
-		public string transformMatrix { get; set; } //CAN BE NULL
-		public uint transformMatrixEditorId { get; set; }
+		public float? scale { get; set; } // Can be null
+		public float? rotate { get; set; } // Can be null
+		public uint? translateX { get; set; } // Can be null
```
zmbq: These two should be grouped.

sqe-dtos/Polygon.cs
```diff
@@ -4,7 +4,10 @@ public class PolygonDTO
 	{
 		public string mask { get; set; }
 		public uint maskEditorId { get; set; }
-		public string transformMatrix { get; set; } //CAN BE NULL
-		public uint transformMatrixEditorId { get; set; }
+		public float? scale { get; set; } // Can be null
```
zmbq: Since scale, rotation and translation seem to be grouped together in numerous places, perhaps we should have a TransormationDTO that has a float? scale, float? rotation and OffsetDTO? translation . So if we happen to add another potential transform, we can add it there and have it everywhere. I wouldn't be surprised if a few months down the road we will add a matrix for the transform (applied after the scale, rotation and translation), for more complicated linear transformations.

sqe-dtos/TextFragment.cs
```diff
@@ -26,6 +26,10 @@ public TextFragmentDataListDTO(List<TextFragmentDataDTO> textFragments)
 			this.textFragments = textFragments;
 		}
 
+		public TextFragmentDataListDTO() : this(null)
```
zmbq: Why not add give the parameter of the first constructor a default null value instead of a whole new constructor?

sqe-http-api/Startup.cs
```diff
@@ -58,6 +61,21 @@ public void ConfigureServices(IServiceCollection services)
...
+			{
+				// A custom compression level makes a huge difference CompressionLevel.Optimal uses level 11,
+				// which is incredibly slow.  A level between 5–7 gives a similarly sized result at a considerably
+				// faster speed. On one test Broti (CompressionLevel)5 compressed a 9.4 MB file to 1.57 MB, gzip
+				// compressed it to 2.45 MB (albeit a little bit faster).
+				options.Level = (CompressionLevel)5;
+			});
+			services.Configure<GzipCompressionProviderOptions>(options =>
+			{
+				options.Level = CompressionLevel.Optimal;
```
zmbq: This may cause stress on the server (not in development, though). Something to keep in mind when we stress test.

zmbq (changes requested):

data-access/Helpers/Exceptions.cs
Bronson-Brown-deVost: Nice catch! I will go through and refactor these.

data-access/ImageRepository.cs
Bronson-Brown-deVost: Let me check this, because I don't really care for how it looks.  UserId is null when an unauthenticated (anonymous) user is making read requests to the API (it is currently a uint? type).  I will check the queries to see if we can simply pass the null straight through and still get the intended results.

data-access/Models/ArtefactModels.cs
Bronson-Brown-deVost: Yeah, I had a long discussion with Ingo about this a few years back. He was not in favor of allowing negative values, by the same token he also wanted the coordinate system to have its origin at top,right, with value increasing as you move from right to left (essentially an RTL coordinate system).  I really feared the inevitable coding and data manipulation errors that would result from that, and opted for the (universally?) standard top,left origin used by both the browser and GIS.  
In any event, I am not opposed to allowing negative values (I would just need to change the type in the database).  Indeed I always feared the issue of moving something a little above or to the left of 0, which in this system would mean moving every single other piece of data down and to the right, ugh.  I think for the user, they will perhaps have rulers in view and I really dislike the idea of having negative values there (but this can be managed browser side).  So, basically, the real issue is computational/representational simplicity; do you prefer to have every scroll begin at point 0,0 and to support negative numbers, or do you prefer to have every scroll begin at point 2147483647,2147483647 (the middle value of a 32 bit unsigned int) and disallow negative numbers?  I could even let you use signed 32 bit integers when communicating with the API and add 2147483647 to the value before inserting into the database.  What is your preference?

data-access/Models/ArtefactModels.cs
Bronson-Brown-deVost: I like the sound of that.  I will leave this as is in the database model (it is used directly by Dapper to serialize the DB query), but in the corresponding DTO I will serialize it to a translate or offset object with x and y properties.  Thanks.

data-access/Models/ArtefactModels.cs
zmbq: Let me see if I can choose a larger font. Here we go:
# NO! 
Top left at (2^31, 2^31) means nobody, ever, is going to be able to read these coordinates and tell you which is to the left of which. People will constantly need to compare 10 digit numbers in their head - and that's a huge mess.

In Israel when they teach you about graphs and the X and Y axes, the X axis goes from left to right (1 is to the right of 0).  So (0,0) as top right is going to confuse every single person looking at the coordinates - Israeli or not. (0,0) as top left is by far the simplest.

As for negatives - I guess it depends on how user friendly we want to be. Having someone move every single piece on the scroll just to make a little room seems very tedious and not at all friendly. We can do this automatically (and generate a huge amount of database updates), but is that really worth it, instead of just allowing negatives? I don't have a strong opinion hear, I'm just wondering. I do feel very strongly about (0,0) being the top left.

data-access/ImageRepository.cs
zmbq: You may just need to change the comment, saying it's not expanded when it's 0 and that's it.

data-access/UserRepository.cs
Bronson-Brown-deVost: I will add a better comment, thanks. Basically userId 1 holds the permissions that apply to an anonymous user (who would naturally have no user id).

data-access/Models/ArtefactModels.cs
Bronson-Brown-deVost: Absolutely agreed on 0,0 at top left.  I also have no strong feeling on unsigned/signed ints, I guess on the programming side signed integers maybe make more sense and would provide more meaningful values.  I don't mind making the adjustment in the database.

sqe-dtos/Polygon.cs
Bronson-Brown-deVost: Yeah, I think that is a good idea.  And I hope we do not add a matrix on top of it.  By the way, I think we need to make sure that the transform operation is properly applied always in the standard SRT (scale, rotate, translate) order so that any GIS functions in the DB/API will match the front end presentation.  A different order of operations can result in a different result (which is one reason I liked the safety of the transform matrix).

sqe-dtos/Artefact.cs
Bronson-Brown-deVost: I will update accordingly.

sqe-dtos/Polygon.cs
Bronson-Brown-deVost: Ok.

data-access/Models/ArtefactModels.cs
zmbq: If Ingo feels strongly about non-negatives, let's keep the uints. If at some point when people use the editor they decide they need negative coordinates, we'll just change it.

Oh. Oh. Oh. That is only relevant for translations of artefacts relative to scrolls. Other translations (master image vs. another image, for instance) may very well include negative numbers (the master image is at (100, 100) and the PAM image is at (150, 50), so the translation is (50, -50).

Bronson-Brown-deVost: Ok.  Indeed, this is only for location of artefacts on the virtual scroll.  Other operations will certainly need negative values as you say.

sqe-dtos/TextFragment.cs
Bronson-Brown-deVost: This is funny.  I think I see the work of some autotooling here, nice catch.  In fact I think you can just pass null to the TextFragmentDataListDTO(List<TextFragmentDataDTO> textFragments) constructor, since `List<T> myVar = null` is valid.

sqe-http-api/Startup.cs
Bronson-Brown-deVost: Yeah, I should add a note about that and also that if we run this on something other than Kestrel (like IIS), then we should perhaps turn it off in favor what functionality that server provides (at least that is what the docs say).

data-access/Models/ArtefactModels.cs
zmbq: That's OK for this model, but I'd like to use the same DTO for all transformations - artefact on scroll or otherwise. So at least the DTO should have ints and not uints.

Bronson-Brown-deVost: Ok, I will switch the DTO to use signed ints.  And yes, if we can re-purpose the DTO, that would be nice.

data-access/ImageRepository.cs
Bronson-Brown-deVost: Looks like I was able to remove all of these, and to pass the uint? directly to the query.  And I found another spot where I was unnecessarily using `?? 1` because of a poorly written query.  That is now fixed too.  Thanks!

data-access/Models/ArtefactModels.cs
zmbq: I saw that you actually subtract 2^31 from the number returned from the database when creating this int. I thought putting the middle point at (2^31, 2^31) was a joke and not something that you actually did - even if only in the database. It is a *horrible horrible idea*. You basically do allow negative numbers, you just encode them in a way that is so unnatural, we may end up as a DailyWTF.

sqe-http-api/Startup.cs
zmbq: In that case, why don't you just remove it entirely for now? We're running the server and client on the same machine anyway (even if it's inside a docker container). Network bandwidth is meaningless. Compress only on an actual server.

zmbq (approved):

PR #25, opened 2019-11-06
Update tests

Bronson-Brown-deVost:
Here is the latest stuff.  I am still updating many of the old tests to the new system, but you might as well merge it now.


.githooks/pre-commit
```diff
@@ -7,28 +7,28 @@ function check_docker_vars
 {
     grep -n '\-[[:space:]]'"${1}"'=[^$]' docker/deploy/docker-compose.yml
     if [[ "$?" == 0 ]]; then 
-        echo "Please remove the custom settings for ${1} from docker/deploy/docker-compose.yml"
-        echo "The line should be:"
-        echo "#- ${1}="
+        printf "\e[31mPlease remove the custom settings for ${1} from docker/deploy/docker-compose.yml\n"
+        printf "The line should be:\n"
+        printf "#- ${1}=\e[0m\n"
```
zmbq, line 12: You shouldn't change docker-compose at all. We have a Powershell script that sets the proper environment variables and runs docker-compose. It's not in the repo since it contains credentials to one of our company mailboxes.

.githooks/pre-commit
```diff
@@ -42,14 +42,29 @@ check_docker_vars "MAILER_EMAIL_SMTP_URL"
...
+        printf "\e[31mThere are errors in the HTTP controllers in 'sqe-api-server/HttpControllers'.\n"
+        printf "Please fix these errors before continuing to commit your changes.\e[0m\n"
+        exit 1
+    fi
+}
+
+## Generating Hub Methods
+echo "Autogenerating realtime Hub methods"
+generate_hubs
+
```
zmbq, line 67: I like the idea of these checks. I hope they won't cause issues with the ordinary workflow. One thing I think of - we try to push whatever it is we have (even when it's not working) when we go home - for backup reasons. We do it in branches that we know don't bother anybody else. Perhaps we should have some naming convention for branches that aren't checked, just for this workflow (it beats setting up a backup service on each and every computer we have)

zmbq (approved):
Since you moved files between directories, almost all the source files in this PR have two entries - the old one deleted, and the new one a newly created file - all with green lines. This means I basically need to read the entire codebase again. This is not something I want to do through the github interface. So I'm approving this PR without doing a real code review process - it will be too time consuming. When we find issues, I'll let you know even after we merge...

PR #28, opened 2019-11-15
Fix batch rois

Bronson-Brown-deVost:
This fixes the bug that POSTs to the batch ROI endpoint would return a successful status and object, but were not actually stored in the database (I forgot to commit the transaction).

A smaller bugfix here is the return data from the text-fragment endpoints, which would send out duplicates of the nextSignInterpretation objects (the object needed a proper Equals and GetHashCode override).

### DIFF
--- a/sqe-database-access/Models/SignInterpretation.cs
+++ b/sqe-database-access/Models/SignInterpretation.cs
@@ -13,7 +13,22 @@ namespace SQE.DatabaseAccess.Models
 
     public class NextSignInterpretation
     {
-        public uint nextSignInterpretationId { get; set; }
-        public uint signSequenceAuthor { get; set; }
+        public uint nextSignInterpretationId { get; }
+        public uint signSequenceAuthor { get; }
+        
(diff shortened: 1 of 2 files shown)


sqe-database-access/Models/SignInterpretation.cs
```diff
@@ -13,7 +13,22 @@ public class SignInterpretation
...
-        public uint signSequenceAuthor { get; set; }
+        public uint nextSignInterpretationId { get; }
+        public uint signSequenceAuthor { get; }
+        
+        // The override for Equals and GetHashCode methods here enable the
+        // HashSet nextSignInterpretations of the SignInterpretation object
+        // to ensure that no duplicate values will be inserted into the set.
+        public override bool Equals(object obj)
+        {
+            return obj is NextSignInterpretation q
```
zmbq, line 24: This `is` operator usage is new to me. I actually don't like it very much, I wonder why they've added it.

zmbq (approved):

PR #30, opened 2019-12-02
patched text retrieval query

Bronson-Brown-deVost:
@zmbq, this PR fixes the issue you noticed where the text order changes after you have saved a few ROI's.

### DIFF
--- a/sqe-database-access/Queries/Text.cs
+++ b/sqe-database-access/Queries/Text.cs
@@ -194,6 +194,9 @@ AS (
 
 		JOIN edition ON edition.edition_id = @EditionId
 	)
+
+SELECT *
+FROM sign_interpretation_ids
 ";
     }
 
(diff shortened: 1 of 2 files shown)


sqe-database-access/Queries/Text.cs
```diff
@@ -9,98 +9,191 @@ public static class GetTextChunk
         ///     @editionId is the Id of the edition the text is to be taken from
         /// </summary>
         public const string GetQuery = @"
```
zmbq, line 11: This is a huge query. I suspect it will cause major performance issues. It's worth running it through a query analyzer (I'm sure MariaDB has one) to see which indices are used and how long it takes to execute.

zmbq (approved):
Tests are failing I see, due to deadlocks.

sqe-database-access/Queries/Text.cs
Bronson-Brown-deVost: Sorry, I made a mistake and did not push the latest commit from this branch up to GitHub.  You will see a cleaner query and, best of all, the tests pass. I did run some query analyses, and the query runs pretty well, but I still have some worries about the time to complete.

I am still thinking of preprocessing the JSON objects for these text queries on each update to the edition, but I want some better usage stats before I get into that.

PR #31, opened 2020-03-05
Staging

Bronson-Brown-deVost:
Sorry this took so long.  I wanted to make some updates to the tests and the documentation and fix a few little bugs here and there.


Bronson-Brown-deVost:
Sorry for the CI fails, this is failing due to the utility we use for reporting code coverage to Coveralls (csmacnz.Coveralls).  It does not seem to work properly with dotnet 3.  I am looking for a fix now.  Otherwise, everything is working properly.

Utilities/GenerateTypescriptInterfaces/Program.cs
```diff
@@ -0,0 +1,330 @@
...
+        {
+            ParseSqeSignalrHub();
+        }
+
+        private static void ParseSqeSignalrHub()
+        {
+            Console.WriteLine("Parsing the SignalR hub and creating corresponding typescript interfaces.");
+
+            // TODO: Can we find a better way to resolve these paths instead of all the backtracking?
+            var projectRoot =
```
zmbq, line 91: Perhaps you can add a command line argument that accepts the root of the project, and fall back to this if there's no root specified?

Bronson-Brown-deVost:
Hey! I fixed it!

Utilities/GenerateTypescriptInterfaces/Program.cs
```diff
@@ -0,0 +1,330 @@
...
+                case "double":
+                case "single":
+                case "uint":
+                case "int":
+                    return "number";
+                case "string":
+                    return "string";
+                case "bool":
+                    return "boolean";
+            }
```
zmbq, line 214: Add a `default` clause that throws an exception, in case we use an unsupported type. For example, DateTime, which we may use, and needs to be a `string` on the Typescript side (there is a Typescript Date class, it's never handled automatically from JSON)

Utilities/GenerateTypescriptInterfaces/SqeHubInterfaceTemplate.txt
```diff
@@ -0,0 +1,23 @@
+import {
```
zmbq, line 1: Where is the tslint disable comment? And a comment saying this file was generated automatically and should not be changed?

sqe-api-server/Helpers/HttpExceptionMiddleware.cs
```diff
@@ -35,22 +36,24 @@ public async Task Invoke(HttpContext context)
                 context.Response.ContentType = "application/json; charset=utf-8";
                 await context.Response.WriteAsync(
                     JsonConvert.SerializeObject(
-                        new ApiExceptionError(nameof(httpException), httpException.Error)
+                        new ApiExceptionError(nameof(httpException), httpException.Error, httpException is IExceptionWithData exceptionWithData ? exceptionWithData.CustomReturnedData : null)
```
zmbq, line 39: Isn't that `(httpException as IExceptionWithData)?.CustomReturnedData` ?

sqe-api-server/Helpers/InputValidation.cs
```diff
@@ -24,5 +42,487 @@ public static bool ValidateTransformMatrix(string transformMatrix)
...
+        /// </summary>
+        /// <param name="wktPolygon">A wkt polygon</param>
+        /// <param name="entityName">The type of object being validated (used in formulating useful exception error messages)</param>
+        /// <param name="fix">Invalid polygons always throw an error, this flag determines whether to try to return a repaired
+        /// version of the polygon with that error</param>
+        /// <returns>A WKT string with the cleaned polygon</returns>
+        public static async Task<string> ValidatePolygonAsync(string wktPolygon, string entityName, bool fix = false)
+        {
+            // Wrap the private validation method in a Task so we can asynchronously call this non-trivial method
+            return await Task.Run(() => _validatePolygon(wktPolygon, entityName, fix));
```
zmbq, line 55: This deserves a very big warning here, and especially in _validatePolygon, which runs in its own thread, and may run several times concurrently.

sqe-api-server/Helpers/InputValidation.cs
```diff
@@ -1,10 +1,28 @@
...
 
 namespace SQE.API.Server.Helpers
 {
     public static class GeometryValidation
     {
+
+        private static readonly WKTReader _wkr = new WKTReader();
+        private static readonly Regex _asymmetricNesting = new Regex(@"\),\s(?!\()");
+        private static readonly Regex _wktSplit = new Regex(@"\)\s?,\s?\(");
+
```
zmbq: Since these are used by multiple threads concurrently, I would definitely not use them as global here. Instead, I would make `GeometryValidation` a non-static class with a static function that performs the validation. This function will instantiate a class instance (it can have a `private` constructor so that nobody else can), which will create its own copy of these, so no object is used twice concurrently.

I'm not sure what happens if you call `Regex.Match` twice concurrently on the same object, and most certainly what happens if a `WKTReader` tries to read two things at once.

sqe-api-server/Helpers/InputValidation.cs
```diff
@@ -24,5 +42,487 @@ public static bool ValidateTransformMatrix(string transformMatrix)
...
+            return fullGeom;
+        }
+
+        /// <summary>
+        /// Fixes a self-intersecting polygon by changing the point of intersection to
+        /// very thin open area. 
+        /// </summary>
+        /// <param name="poly">The self-intersecting polygon to fix</param>
+        /// <returns>A repaired version of the self-intersecting polygon</returns>
+        private static Polygon _repairSelfInterectingPolygon(Geometry poly)
```
zmbq, line 198: Typo - you forgot an s and repair a self interecting polygon

sqe-api-server/Helpers/InputValidation.cs
```diff
@@ -24,5 +42,487 @@ public static bool ValidateTransformMatrix(string transformMatrix)
...
+
+
+            if (discretePolys[0].IsValid)
+                return (Polygon)discretePolys[0];
+
+            // Probably the polygon is invalid due to self intersection
+            // Let's grab every outer line in the polygon
+            var coords = poly.Boundary.Coordinates;
+            var lines = coords.Select(
+                (t, i) => new LineString(new[] { t, coords[i + 1 == coords.Length ? 0 : i + 1] })
```
zmbq, line 214: I would write it as `coords[(i + 1) % coords.Length]`;

sqe-api-server/Helpers/InputValidation.cs
```diff
@@ -24,5 +42,487 @@ public static bool ValidateTransformMatrix(string transformMatrix)
...
+
+            // Now group the lines into linestrings that start and stop at the points of intersection
+            var intersectedLines = new List<LineString>();
+            var points = new List<Coordinate>();
+            foreach (var line in lines)
+            {
+                if (!points.Any())
+                    points.Add(line.Coordinates[0]);
+
+                foreach (var compLine in lines)
```
zmbq, line 225: This is O(N^2), which may be a lot in large polygons.

sqe-api-server/Helpers/InputValidation.cs
```diff
@@ -24,5 +42,487 @@ public static bool ValidateTransformMatrix(string transformMatrix)
...
+                {
+                    if (line.Coordinates.Contains(compLine.Coordinates[0])
```
zmbq, line 227: Are you sure this test is correct? What if we have an intersection that is represented by one point that is a corner of four lines?

sqe-api-server/Helpers/InputValidation.cs
```diff
@@ -24,5 +42,487 @@ public static bool ValidateTransformMatrix(string transformMatrix)
...
+                    intersectedLines.Add(new LineString(points.ToArray()));
+                    points = new List<Coordinate>();
+                    points.Add(point.Coordinate);
+                }
+
+                points.Add(line.Coordinates[1]);
+                if (lines.Last().ToText() == line.ToText())
+                    intersectedLines.Add(new LineString(points.ToArray()));
+            }
+
```
zmbq, line 243: This function is actually impossible to follow and understand. It is the kind of calculation that I think deserves long comments with a concrete example of polygons.

sqe-api-server/HttpControllers/EditionController.cs
```diff
@@ -25,27 +25,43 @@ public EditionController(IEditionService editionService, IUserService userServic
...
         {
-            return await _editionService.AddEditionEditor(
+            return await _editionService.RequestNewEditionEditor(
                 await _userService.GetCurrentUserObjectAsync(editionId, admin: true),
                 payload
             );
         }
 
+        /// <summary>
+        ///     Confirma addition of an editor to the specified edition
```
zmbq, line 39: Typo, or perhaps you meant to write in an Italian accent.

sqe-api-server/HttpControllers/UtilController.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        /// <summary>
+        ///     Checks a WKT polygon to ensure validity. If the polygon is invalid,
+        ///     it attempts to construct a valid polygon that matches the original
+        ///     as closely as possible.
+        /// </summary>
+        /// <param name="payload">JSON object with the WKT polygon to validate</param>
+        [HttpPost("v1/[controller]s/validate-wkt")]
+        public async Task<ActionResult> RequestAddEditionEditor([FromBody] WktPolygonDTO payload)
```
zmbq, line 27: I think you forgot to rename the function after copying it...

sqe-api-server/RealtimeHubs/ArtefactHub.cs
```diff
@@ -13,6 +13,12 @@
 using SQE.API.Server.Services;
 using Microsoft.AspNetCore.SignalR;
 
+using SQE.DatabaseAccess.Helpers;
```
zmbq, line 16: Didn't we say that all the Hub files shouldn't be part of the repository, as they are created during the build process?

sqe-api-server/RealtimeHubs/SubscriptionHub.cs
```diff
@@ -29,29 +54,36 @@ public partial class MainHub
...
         {
-            var user = await _userService.GetCurrentUserObjectAsync(editionId);
+            try
+            {
+                var user = await _userService.GetCurrentUserObjectAsync(editionId);
 
-            if (!user.MayRead)
-                throw new StandardExceptions.NoReadPermissionsException(user);
+                if (!user.MayRead)
+                    throw new StandardExceptions.NoReadPermissionsException(user);
```
zmbq, line 62: Isn't it better to throw an exception saying the edition does not exist? If the user has no permissions to it, it should be as if there's no edition at all.

sqe-api-server/Services/ArtefactService.cs
```diff
@@ -76,6 +86,14 @@ public ArtefactService(IArtefactRepository artefactRepository, IHubContext<MainH
                 );
             }
 
+            var wkr = new WKTReader();
+            var wkw = new WKTWriter();
+            artefacts.artefacts = artefacts.artefacts.Select(x =>
```
zmbq: I'm not sure what this does. I think a comment can be helpful.

sqe-api-server/Services/EditionService.cs
```diff
@@ -113,8 +146,10 @@ public async Task<EditionListDTO> ListEditionsAsync(uint? userId)
             var updatedEdition = EditionModelToDTO(editions.First(x => x.EditionId == editionUser.EditionId));
             // Broadcast the change to all subscribers of the editionId. Exclude the client (not the user), which
             // made the request, that client directly received the response.
-            await _hubContext.Clients.GroupExcept(editionUser.EditionId.ToString(), clientId)
-                .SendAsync("updateEdition", updatedEdition);
+            var editionUsers = await _editionRepo.GetEditionEditorUserIds(editionUser);
+            foreach (var userId in editionUsers)
+                await _hubContext.Clients.GroupExcept($"user-{userId.ToString()}", clientId)
```
zmbq, line 151: The string `$"user-{userId.ToString()}"` appears way too many times. I would add a helper method to the hub context (or somewhere else) that gets a userId as input, and returns what `GroupExcept` returns. This can eliminate bugs that are going to be very hard to catch (if you mistype the string once, in one function, it can result in a bug that happens once a month, depending on how infrequent that function is)

sqe-api-server/Services/EditionService.cs
```diff
@@ -240,43 +284,179 @@ await _hubContext.Clients.GroupExcept(editionUser.EditionId.ToString(), clientId
...
-                .SendAsync("updateEditionEditor", updatedEditorDTO);
+            var editionUsers = await _editionRepo.GetEditionEditorUserIds(editionUser);
+            foreach (var editionUserId in editionUsers)
+                await _hubContext.Clients.GroupExcept($"user-{editionUserId.ToString()}", clientId)
+                    .CreatedEditor(updatedEditorDTO);
+
             return updatedEditorDTO;
         }
 
+        public async Task<EditionScriptCollectionDTO> GetEditionScriptCollection(EditionUserInfo editionUser)
```
zmbq, line 402: I can guess what this does, but I think it needs a comment.

sqe-api-server/Services/EditionService.cs
```diff
@@ -240,43 +284,179 @@ await _hubContext.Clients.GroupExcept(editionUser.EditionId.ToString(), clientId
...
+        {
+            var letters = await _editionRepo.GetEditionScriptCollection(editionUser);
+            var lettersSorted = letters.GroupBy(x => x.Id).ToList();
+            var wkbr = new WKBReader();
+            var wkr = new WKTReader();
+            var wkw = new WKTWriter();
+            return new EditionScriptCollectionDTO()
```
zmbq, line 409: I would break the following statement into several statements, and maybe put the inner select in its own function. I'm not sure how easy this is going to be to debug.

sqe-api-server/Services/TextService.cs
```diff
@@ -104,16 +121,58 @@ public async Task<TextFragmentDataListDTO> GetFragmentDataAsync(EditionUserInfo
...
+        /// <summary>
+        ///     Serialize a TextEdition and the list of its editors to a TextEditionDTO.
+        /// </summary>
+        /// <param name="ed">Text edition to be serialized</param>
+        /// <param name="editors">List of edition editors</param>
+        /// <returns>A TextEditionDTO</returns>
         private static TextEditionDTO _textEditionToDTO(TextEdition ed, List<EditorInfo> editors)
         {
-            var editorList = new Dictionary<uint, EditorDTO>();
+            var editorList = new Dictionary<string, EditorDTO>();
```
zmbq, line 172: Why change from uint to string here?

sqe-database-access/EditionRepository.cs
```diff
@@ -393,29 +415,101 @@ public async Task<string> GetDeleteToken(EditionUserInfo editionUser)
...
+                        new List<string> { "email" }
+                    ),
+                    new { Email = editorEmail }
+                );
+
+                // TODO: Use the database user email token system rather than a JWT for this (see commented code below.
+                // That way, we can query for "open" editor requests. Must wait to update the database before we can
+                // implement this functionality.
+                var secret = _config.GetSection("AppSettings").GetSection("Secret").Value;
+                editorInfo.Token = new JwtBuilder()
```
zmbq, line 432: Storing all the permission information in the token is nice, but I guess this will need to be changed if we want to list all pending invitations - there will be a database table for pending invitations, and the token will contain just the ID of that table, right?

sqe-database-access/Models/UserModels.cs
```diff
@@ -36,6 +36,17 @@ public class UserEditionPermissions
         public bool Locked { get; set; }
     }
 
+    public class AddEditorJWT
```
zmbq, line 39: If this is stored in the token, and nowhere else, I don't think it belongs in the database-access assembly. It's going to be changed into an actual model that is going to be saved in the database, so this is just a bit of nitpicking.

sqe-realtime-hub-builder/Program.cs
```diff
@@ -69,24 +94,34 @@ private static void ParseSqeHttpControllers()
...
             var files = dir.GetFiles("*.cs");
             var classFields = new List<ControllerField>();
+            var hubInterfaces = new List<string>();
 
             // Delete any existing Realtime Hubs
             Console.WriteLine($"Deleting any existing realtime hub methods from {hubFolder}.");
             var folderInfo = new DirectoryInfo(hubFolder);
-            foreach (var file in folderInfo.GetFiles()) file.Delete();
+            foreach (var file in folderInfo.GetFiles())
+                if (file.Name != "HubInterface.cs")
```
zmbq, line 103: I'm a little worried about a directory that includes both auto generated *and* manually written code files. Perhaps the auto generated files should all go into the `AutoGenerated` subdirectory (they can still use the same namespace, so nobody else notices it)?

zmbq (changes requested):
This looks good! I didn't go into the repository code and didn't try to understand the SELECTs as well as the new entities you're handling (letter shapes, for instance).

There is one major problem, though, which is why I don't approve this request - I think the polygon validation code may break, badly, when called several times concurrently. I've left some remarks with suggestions how to handle that.

I'm also not sure the algorithm works in all cases, but we don't need to handle all cases - just those that happen when people draw boundaries around letters and fragments.

sqe-database-access/Models/InterpretationAttribute.cs
```diff
@@ -8,6 +8,20 @@ public class CharAttribute
...
+        // to ensure that no duplicate values will be inserted into the set.
+        public override bool Equals(object obj)
+        {
+            return obj is CharAttribute q 
+                   && q.interpretationAttributeId == this.interpretationAttributeId;
+        }
+
+        public override int GetHashCode()
+        {
+            return unchecked((int) this.interpretationAttributeId);
```
zmbq: This had me going to read about `unchecked`, since I've never used it before. Turns out `(int)u` is enough, no need for unchecked. No need to change the code, just something that's useful to know:
https://stackoverflow.com/a/1131851/871910

zmbq (approved):

PR #32, opened 2020-03-31
Invitation tokens

Bronson-Brown-deVost:
All tests pass but the coverage is still not as good as I would like. Perhaps you should give a little real world testing first.


zmbq:
There are a few bugs, documented on Redmine. I'll check the pull request once the bugs have been ironed out.

Bronson-Brown-deVost:
Thank you for testing this out a bit.  I should be able to get to these bugs this afternoon.

Bronson

qwb-to-sqe/Data/QWBSign.cs
```diff
@@ -0,0 +1,94 @@
+using System;
```
zmbq, line 1: I am not looking at the qwb-to-sqe project at all. I can if you want me to, but then you'll need to explain a little about the QWB database.

sqe-api-server/Helpers/InputValidation.cs
```diff
@@ -70,39 +70,32 @@ public static async Task<string> ValidatePolygonAsync(string wktPolygon, string
         private static string _validatePolygon(string wktPolygon, string entityName, bool fix)
```
zmbq, line 70: This is not code I can review properly on github. It's too long. However, we've seen it work OK for most of the cases, so I'm going to approve this. If you want, we can have an online meeting where you describe the algorithm and the code. It shouldn't block this PR, though.

sqe-api-server/Services/TextService.cs
```diff
@@ -197,69 +198,69 @@ private static TextEditionDTO _textEditionToDTO(TextEdition ed, List<EditorInfo>
                 textFragments = ed.fragments.Select(
                         x => new TextFragmentDTO
                         {
-                            textFragmentId = x.textFragmentId,
-                            textFragmentName = x.textFragmentName,
-                            editorId = x.textFragmentAuthor,
+                            textFragmentId = x.TextFragmentId.GetValueOrDefault(),
```
zmbq, line 201: You use a lot of nullables, both here and in the sign interpretation entities in the previous file. Why can all these be null? When is the X translation null? How can the text fragment ID of a text fragment be null?

sqe-database-access/Helpers/DatabaseWriter.cs
```diff
@@ -153,61 +155,75 @@ public DatabaseWriter(IConfiguration config) : base(config)
...
+            // If there is no currently running transaction, start a new one
+            if (Transaction.Current == null)
+            {
+                return await DatabaseCommunicationRetryPolicy.ExecuteRetry(
+                    async () =>
                     {
-                        foreach (var mutationRequest in mutationRequests)
+                        // Grab a transaction scope, we roll back all changes if any transactions fail
+                        // I could limit the transaction scope to each individual mutation request,
+                        // but I fear the multiple requests may be dependent upon each other (i.e., all or nothing).
```
zmbq, line 166: Doing everything in one transaction is a good idea.

zmbq (approved):

sqe-api-server/Services/TextService.cs
Bronson-Brown-deVost: For posterity, this was Ingo's code.  He wanted to use nullable values to limit the proliferation of objects (this makes the object quite multipurpose).  I'll keep an eye on that though in case it causes more problems than it solves in the end.

PR #34, opened 2020-06-04
Gis updates

Bronson-Brown-deVost:
Here is the latest work on the API.  

1. We have a small fix to the query which gathers the text transcriptions.  
2. There is an added GIS endpoint to get letter ROIs by line. 
3. We have added support for version 0.17.0 of the database, which allows a signed int value for the artefact zIndex.  
4. There are some bugfixes, enhancements, and added documentation for the artefact update functionality.  
5. There is now an endpoint for batch updates of artefact position.


Bronson-Brown-deVost:
Oh and I see in the commits that this also changes to use the new rust repair poly module.  Probably you see also that a lot of the polygon tests have been removed from UtilTests.cs, they are part of the rust library now, specifically: https://gitlab.com/bronsonbdevost/rust-geo-repair-polygon/-/blob/master/src/repair.rs

Qumran.sln.DotSettings.user
```diff
@@ -11,4 +11,7 @@
 	<s:String x:Key="/Default/Environment/AssemblyExplorer/XmlDocument/@EntryValue">&lt;AssemblyExplorer&gt;
   &lt;Assembly Path="/Users/bronson/.nuget/packages/nettopologysuite/2.0.0/lib/netstandard2.0/NetTopologySuite.dll" /&gt;
 &lt;/AssemblyExplorer&gt;</s:String>
+	<s:String x:Key="/Default/Environment/UnitTesting/UnitTestSessionStore/Sessions/=91b9dfa2_002D0c15_002D4a3d_002Da870_002D361628bfb78c/@EntryIndexedValue">&lt;SessionState ContinuousTestingIsOn="False" ContinuousTestingMode="0" FrameworkVersion="{x:Null}" IsLocked="False" Name="All tests from &amp;lt;sqe-api-test&amp;gt;" PlatformMonoPreference="{x:Null}" PlatformType="{x:Null}" xmlns="urn:schemas-jetbrains-com:jetbrains-ut-session" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"&gt;
```
Bronson-Brown-deVost: @zmbq, do you know what this is? It must be added by my IDE, but it looks unnecessary. I will try deleting it if you think that is ok.

zmbq: I have no idea what this is. I never look at `.user` files. They shouldn't be in git, either.

sqe-api-server/Helpers/ArtefactDTOTransformer.cs
```diff
@@ -19,18 +19,21 @@ public static ArtefactDTO QueryArtefactToArtefactDTO(ArtefactModel artefact, uin
...
-                                y = artefact.TranslateY.Value
-                            }
-                            : null
-                    },
+                    transformation = artefact.Scale.HasValue
+                        ? new TransformationDTO
+                        {
+                            scale = artefact.Scale.Value,
+                            rotate = artefact.Rotate.Value,
+                            zIndex = artefact.ZIndex.Value,
```
zmbq, line 27: Oh, you've moved zIndex into the TransformationDTO! It most certainly makes sense to put it here. It will make things clearer in the frontend.

sqe-api-server/Helpers/InputValidation.cs
```diff
@@ -18,6 +13,43 @@ namespace SQE.API.Server.Helpers
...
+        [DllImport(
+            "geo_repair_polygon",
+            CharSet = CharSet.Ansi,
+            CallingConvention = CallingConvention.Cdecl
+        )]
+        private static extern unsafe IntPtr repair_wkt(IntPtr wkt);
+
+        [DllImport(
+            "geo_repair_polygon",
+            CharSet = CharSet.Ansi,
```
zmbq, line 48: Isn't RUST UTF-8 by default? It doesn't make any difference for the wkt polygons we move around, but still.

Qumran.sln.DotSettings.user
Bronson-Brown-deVost: Thanks, I will put *.DotSettings.user into gitignore then. Thanks.

sqe-api-server/Helpers/InputValidation.cs
```diff
@@ -74,14 +106,63 @@ private static string _validatePolygon(string wktPolygon, string entityName, boo
...
+                // Try repairing the binary version of the polygon
+                var wkb_in = polygon.AsBinary(); // Get the binary data
+                // Instantiate the variable to process the response from the FFI
+                var bin = new BinaryData();
+                var wkbData = new Span<byte>();
+
+                // Reading and writing pointers is unsafe
+                try
+                {
+                    unsafe
```
zmbq: I don't like unsafe code one bit. It limits your ability to run it under various conditions (a web server is one of them), and may require you to perform some not-so-nice configuration changes.

Also, there's no reason to use unsafe code here. Use MarshalAs(UnmanagedType.LPArray). See [here](https://docs.microsoft.com/en-us/dotnet/framework/interop/default-marshaling-for-arrays) for an example of how to pass variable length arrays as in/out parameters to external code.

I would really prefer it if we don't push unsafe code to development unless we absolutely must (in which case, you should put a very clear comment as to why)

Qumran.sln.DotSettings.user
Bronson-Brown-deVost: Hmmm. *.user is already in the .gitignore file, I wonder how this got in then.

Bronson-Brown-deVost: OK, I just deleted the file and there seems to be no problem.

sqe-api-server/RealtimeHubs/ArtefactHub.cs
```diff
@@ -168,5 +184,26 @@ public async Task<ArtefactDTO> PutV1EditionsEditionIdArtefactsArtefactId(uint ed
...
 
 
+        /// <summary>
+        ///     Updates the positional data for a batch of artefacts
+        /// </summary>
+        /// <param name="editionId">Unique Id of the desired edition</param>
+        /// <param name="payload">A BatchUpdateArtefactTransformDTO with a list of the desired updates</param>
+        /// <returns></returns>
+        [Authorize]
+        public async Task<BatchUpdatedArtefactTransformDTO> PostV1EditionsEditionIdArtefactsBatchTransformation(uint editionId, BatchUpdateArtefactTransformDTO payload)
```
zmbq, line 194: I don't think this is a good notification method. When I'm a client listening for artefact updates, I really don't care whether the update was done in a batch or not, I just want to listen to artefact updates. If I need to listen to both notifications (batch update and single update) - we have a non-friendly interface. And if I can listen to either one, it means we're sending the same event (artefact 5 updated) multiple times. I suggest we drop all batch update notifications - a batch update notification is just a series of single notifications.

Since SingalR keeps an open TCP/IP connection, we will incur almost no network overhead by sending 10 small messages instead of one large one. We may use a little more bandwidth, but all the messages will probably fit inside one TCP/IP packet anyway - we won't notice any performance degradation.

sqe-api-server/Services/EditionService.cs
```diff
@@ -508,35 +501,108 @@ public async Task<EditionScriptCollectionDTO> GetEditionScriptCollection(Edition
...
 
+        // TODO: we need to gather also the editor ID's
+        public async Task<EditionScriptLinesDTO> GetEditionScriptLines(EditionUserInfo editionUser)
+        {
+            var wkw = new WKTWriter();
+            var wbr = new WKBReader();
+            var results = await _editionRepo.GetEditionScriptLines(editionUser);
+            return new EditionScriptLinesDTO()
+            {
+                textFragments = results.Select(a => new ScriptTextFragmentDTO()
```
zmbq: This seems to be overly nested to be readable. The finale, with all those ).ToList()).ToList()... is extremely confusing. Can this be broken into smaller functions - one that returns the characters, one that returns the lines, and maybe one that returns the transformation as well?

sqe-api-server/sqe-api-server.csproj
```diff
@@ -7,10 +7,12 @@
     <RootNamespace>SQE.API.Server</RootNamespace>
     <Configurations>Debug;Release</Configurations>
     <Platforms>AnyCPU</Platforms>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
zmbq: This right here may make deploying on some servers problematic (although if we control the servers, it won't be problematic).

sqe-database-access/Queries/SpatialQueries.cs
```diff
@@ -0,0 +1,81 @@
+namespace SQE.DatabaseAccess.Queries
+{
+    public static class EditionScriptLines
+    {
+        public static string GetQuery = @"
+SELECT DISTINCT text_fragment_data.text_fragment_id AS TextFragmentId,
```
zmbq, line 6: I'm sooo happy you guys are writing the SQL queries...

sqe-dto/Transformation.cs
```diff
@@ -4,8 +4,11 @@ namespace SQE.API.DTO
 {
     public class TransformationDTO
     {
-        public float? scale { get; set; }
-        public float? rotate { get; set; }
+        [Range(0, 99.9999)]
+        public float scale { get; set; }
```
zmbq: Does someone enforce the Range? Cool! Anyway, I would make the Range for Scale run from 0.01 to 100, which means we can make it 100 times smaller or 100 times larger.

sqe-dto/Transformation.cs
```diff
@@ -4,8 +4,11 @@ namespace SQE.API.DTO
...
+        [Range(0, 9999.99)]
```
zmbq: Why not limit it from 0 to 360?

zmbq (changes requested):
The two things that are most important in my opinion is the batch update notification followed by the unsafe code (unsafe is second, because it doesn't affect our API, just the ability of some of the people who will install the server themselves to install it).

sqe-api-server/Helpers/InputValidation.cs
Bronson-Brown-deVost: I am checking through this again now.  I agree with you about the unsafe code.  It may have had something to do with freeing the memory, but I am going back now and testing out the approaches without unsafe.  Then we will either get rid of unsafe (I hope to do that) or we will have a nice detailed note as to why we had to use unsafe.  Thanks.

sqe-api-server/RealtimeHubs/ArtefactHub.cs
Bronson-Brown-deVost: This is a great suggestion!  I will make the change.

sqe-api-server/Services/EditionService.cs
Bronson-Brown-deVost: Sure, I am happy to extract those processes into methods.  Who knows, some may even be reusable!

sqe-dto/Transformation.cs
Bronson-Brown-deVost: Yeah, dotnet does the enforcement when it marshalls the HTTP request JSON body into a c# object.  It sends a 400 error when these constraints aren't met.  The number ranges just come straight from the underlying SQL data type: scale is a decimal(6,4), so two digits before the decimal point and 4 after (3 bytes total storage space), and rotate is decimal(6,2), 4 digits before the decimal and two digits after (also 3 bytes).  I had to use decimal in the DB due to some problems with floating point equality comparisons and the whole database writing system.  I can set these however you want, and using 0.1 instead of 0 is a great idea (sadly, I can't give you 100 though).  

What should I do to help you know that floating point values with many digits after the decimal will be truncated?  I considered precision to the millionth sufficient for scale and to the hundredth sufficient for degrees of rotation (I doubt we even have real precision in the SVG browser representation at these levels).

sqe-dto/Transformation.cs
Bronson-Brown-deVost: Sure I can do that.

sqe-api-server/Helpers/InputValidation.cs
Bronson-Brown-deVost: Yeah, but this travels over the C ABI, so ASCII is expected.  Any unicode characters would be invalid anyway, so it is fine to let it fail.

Bronson-Brown-deVost:
OK I made the edits you asked for.  This was very helpful.

1. No more unsafe code (I don't think there is any difference in performance either). Yay!

2. Batch edits to artefact position are treated like a bunch of single edits to the artefact.  Really good call on that one.

3. For conversion to EditionScriptLinesDTO, I broke things down a bit.  In fact I an trying out a new approach which is to create an extension method ToDTO() for the type returned from the repository request.  Check out `sqe-api-server/Serialization/GisConversion.cs`and let me know what you think.  It seems to me that doing things this way would organize our code better and result in a lot of re-usability.

4. I enhanced the validation system for the TransformationDTO.  Firstly, I use a Decimal type which matches the database type. Secondly, I added a custom ValidationAttribute that makes absolutely sure the input matches the database requirements.  Please check that the Javascript Number type can really handle this (e.g., that a Number like 1.1 really is exactly 1.1 in the serialized JSON and not 1.10000002 or whatever). Thirdly, I limit scale to 0.1–99.9999 and rotate to 0–360.  Also, you get a clear 400 message telling you exactly what you did wrong when you violate the validation conditions.

Thanks again.

zmbq (approved):

PR #35, opened 2020-06-08
updated artefact DTOs

Bronson-Brown-deVost:
@Itay here are the updates to the ArtefactDTO system you wanted.  Let me know if I missed something.


sqe-api-server/Serialization/ArtefactConversion.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            return new ArtefactDTO
+            {
+                id = artefact.ArtefactId,
+                editionId = editionId,
+                imageId = artefact.ImageId,
+                artefactDataEditorId = artefact.ArtefactDataEditorId,
+                mask = artefact.Mask,
+                artefactMaskEditorId = artefact.MaskEditorId,
+                isPlaced = artefact.Rotate.HasValue,
```
zmbq: I think `isPlaced` should be true if the translation is defined. We can have a rotation even if the artefact hasn't been placed (if we rotate it in the artefact editor)

sqe-dto/Transformation.cs
```diff
@@ -3,24 +3,33 @@
...
         public uint zIndex { get; set; }
+
+        [Required]
         public TranslateDTO translate { get; set; }
     }
 
     public class TranslateDTO
     {
-        [Required] public uint x { get; set; }
+        [Required]
```
zmbq: A required nullable?

Bronson-Brown-deVost: Again, you are right.  That is a mistake.  Thanks.

sqe-api-server/Serialization/ArtefactConversion.cs
Bronson-Brown-deVost: Oops! you are right.

zmbq (approved):

PR #38, opened 2020-07-09
better error mesage for empty string in EditionCopyDTO

Bronson-Brown-deVost:
Easiest PR ever...


ConsoleApp1/Program.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
```
zmbq, line 1: I think this got into this PR by mistake, when you created another project.

data-to-sqe/Program.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
```
zmbq, line 1: This also seems strange.

zmbq (commented):
I'm not sure I should review this PR, I have no idea what it's supposed to do (probably import QWB data into SQE?)

PR #40, opened 2020-07-27
artefact batch position update fix, update to int for translateX/Y an…

Bronson-Brown-deVost:
Here is the patch I promised. I had to update the database as well, since the problem had to do with us switching the translation x/y from uint to int. I updated the tests to check for this sort of error as well.

### DIFF
--- a/sqe-dto/Transformation.cs
+++ b/sqe-dto/Transformation.cs
@@ -20,13 +20,12 @@ namespace SQE.API.DTO
         [Required]
         public int zIndex { get; set; }
 
-        [Required]
         public TranslateDTO translate { get; set; }
     }
 
     public class TranslateDTO
     {
-        public int? x { get; set; }
-        public int? y { get; set; }
+        public int x { get; set; }
+        public int y { get; set; }
     }
 }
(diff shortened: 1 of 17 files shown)


sqe-dto/Transformation.cs
```diff
@@ -18,15 +18,15 @@ public class PlacementDTO
...
 
         [Required]
         public TranslateDTO translate { get; set; }
     }
 
     public class TranslateDTO
     {
-        public uint? x { get; set; }
-        public uint? y { get; set; }
+        public int? x { get; set; }
```
zmbq: Why are these nullable if we have isPlaced? Under what circumstances are these going to be null?

zmbq (approved):
I'm going to approve this PR so that Shaindel can keep working on this, but I think translateX and Y should not be nullable in the DTO (and probably not in the database, either, but that's less my field here)

zmbq:
I can't merge since the Tcoverage report is unavailable. So perhaps we should wait until we conclude the nullable X/Y issue, too.

Bronson-Brown-deVost:
CI is giving us some trouble here too (hence no report to coveralls). I am getting a random error that I see others are also experiencing intermittently on TravisCI. I am just now fixing a deadlock issue that keeps popping up in the tests. We'll see if that helps.

zmbq:
Perhaps this is, again, because of the memory demands of the backend and database, that Travis CI can't handle?

zmbq (approved):

zmbq:
No coverage yet :-(

PR #41, opened 2020-07-30
Image ref match updates

Bronson-Brown-deVost:
This PR includes the results of some debugging of various performance issues and updates for system administration and cataloguing data.

**Performance and tests:**
The copy edition method has been diagnosed and sped up by a factor of 4 (see the note to CopyEditionSlowAsync in EditionRepository.cs.  We must certainly remove CopyEditionSlowAsync, but I wanted to spend a little testing the new function to make sure no new problem crops up.

The mutate queries in MutateQueries.cs have been updated, it turns out they were far less performant than possible due to the query syntax, they no longer use list comparison. In some cases the performance is up to 50 times faster (<.05 ms vs. nearly 1.5 seconds).

The tests have been updated in many places and due to the above changes, it seems that deadlock has been sharply curtailed.

**System administration**
The database now contains the concept of user roles. These are system roles that enable the development of API accessible maintenance and update functionality. This will certainly include the functionality to add images to the system (something users are not allowed to do because of copyright issues) as well as maintaining some of the reference cataloguing data, which is provided by the institutions and also not user editable. It might also include some helps for user management among various other conveniences for system administration. 

The API can now retrieve the user role information as part of the UserInfo object that accompanies all authenticated API requests and use it to ensure the user has the role necessary for any given request.

**Catalogue Endpoints**
I have added several further endpoints for making requests for cataloguing information from the system (specifically the information about imaged-object to text-fragment matching catalogues).  It also has three new endpoints for updating that catalogue information; these make use of the new system roles functionality to authorize requests. We will be using these endpoints over the coming months to correct the invalid catalogue data in our system.


sqe-api-server/Helpers/InputValidation.cs
```diff
@@ -106,14 +109,17 @@ private static string _validatePolygon(string wktPolygon, string entityName, boo
...
             try
             {
-                polygon = wkr.Read(wktPolygon);
+                var polygon = wkr.Read(wktPolygon);
                 // If it is valid, return it
                 if (polygon.IsValid)
-                    return polygon.Normalized().ToString(); // Always normalize the polygon
+                {
+                    // Remove any completely unnecessary points
+                    var simplifier = new DouglasPeuckerSimplifier(polygon) { DistanceTolerance = 0 };
```
zmbq, line 120: Are you using this instead of the code you wrote a couple of months ago?

sqe-api-server/Serialization/ImagedObjectConversion.cs
```diff
@@ -0,0 +1,67 @@
...
+            };
+        }
+
+        public static SimpleImageDTO ToDTO(this ImagedObjectImage image)
+        {
+            return new SimpleImageDTO()
+            {
+                catalogNumber = image.image_catalog_id,
+                id = image.image_catalog_id,
+                lightingDirection = image.img_type switch
```
zmbq, line 34: I'm really not crazy about all those new C# editions, that just seem, well, badly integrated into the language. The whole addition of the `switch` "operator" - not crazy about it. Don't change the code, I just had to write it.

sqe-api-server/Services/ImageService.cs
```diff
@@ -13,6 +14,7 @@ public interface IImageService
     {
         ImageDTO ImageToDTO(Image model);
         Task<ImageInstitutionListDTO> GetImageInstitutionsAsync();
+        Task<InstitutionalImageListDTO> GetInstitutionImagesAsync(string institution);
```
zmbq, line 17: Don't we need an institution ID and an institution DTO with maybe more than just a name, and some institution metadata?

sqe-database-access/Models/UserModels.cs
```diff
@@ -38,19 +39,42 @@ public class UserEditionPermissions
...
+        USER_ADMIN = 4
+    }
+
+    public class UserInfo
     {
         private readonly IUserRepository _userRepo;
         public readonly uint? userId;
 
-        public EditionUserInfo(uint? userId, uint editionId, IUserRepository userRepository)
+        public UserInfo(uint? userId, uint? editionId, IUserRepository userRepository)
```
zmbq, line 68: I don't remember - why does that `UserInfo` object contain an edition ID?

sqe-database-access/Queries/Image.cs
```diff
@@ -121,6 +147,19 @@ internal class Result
         }
     }
 
+    internal static class InstitutionImagesQuery
+    {
+        public const string GetQuery = @"
+SELECT image_catalog.object_id AS Name, 
+       CONCAT(image_urls.url, SQE_image.filename, '/full/150,/0/', image_urls.suffix) AS Thumbnail,
```
zmbq, line 154: Hmmm, do you really want to keep the IIIF protocol implementation in the SQL queries? I'm actually not sure whether we want to keep it in the backend *at all*. Why not let the API clients decide which image size they want?

sqe-dto/Catalogue.cs
```diff
@@ -0,0 +1,55 @@
...
+
+        [JsonConverter(typeof(JsonStringEnumConverter))]
+        public SideDesignation catalogSide { get; set; }
+        public string imagedObjectId { get; set; }
+        public uint manuscriptId { get; set; }
+        public string editionName { get; set; }
+        public string editionVolume { get; set; }
+        public string editionLocation1 { get; set; }
+        public string editionLocation2 { get; set; }
+        [JsonConverter(typeof(JsonStringEnumConverter))]
```
zmbq, line 22: I think we need a better Typescript generator. These enums should be converted into union string types in Typescript, and that type should be used on the Typescript side.

zmbq (approved):

sqe-api-server/Helpers/InputValidation.cs
Bronson-Brown-deVost: This code still uses the rust code, but it also makes a pass to remove any unnecessary point (e.g., a line with three point 0 0, 10 0, 20 0 gets simplified to two points 0 0, 20 0).  Its not really part of the polygon validation, just a little something to save some space since polygons from the front end often have unnecessary points.

sqe-api-server/Services/ImageService.cs
Bronson-Brown-deVost: I'll make a note and see what other metadata we might want to add for the institution. It might be nice to have a little more info.

sqe-database-access/Queries/Image.cs
Bronson-Brown-deVost: I have no problems with changing this.  I also don't love building this thumbnail address server side.  I'll give it some thought, maybe we should just give the bare IIIF URL address (without the sizing stuff).

sqe-dto/Catalogue.cs
Bronson-Brown-deVost: Agreed.  I'll tackle it this week.

PR #42, opened 2020-07-31
Formatting cleanup

Bronson-Brown-deVost:
Don't merge this until Image ref match updates #41 has been merged.  This PR is nothing but formatting cleanup, renaming of unclear files, and the removal of a bespoke ToPascalCase function in favor of the CaseExtensions nuget package used elsewhere in the project.

So, basically there is really nothing to check.


data-to-sqe/qwb-to-sqe/Common/SshDatabase.cs
```diff
@@ -8,11 +8,9 @@ namespace qwb_to_sqe.Common
 {
     public class SshDatabase
     {
-        private SshClient _sshClient;
-        public MySqlConnection connection { get; private set; }
+        private readonly IConfiguration _configuration;
         private ForwardedPortLocal _port;
-        private IConfiguration _configuration;
-
+        private SshClient _sshClient;
```
zmbq, line 13: This may mean opening a connection takes longer. I think we should look at the entire "connecting to database" part of the backend - it gives you a lot of issues in tests, which shouldn't really be the case. Also, is there connection pooling?

sqe-api-server/Helpers/InputValidation.cs
```diff
@@ -71,32 +63,38 @@ public static bool ValidateTransformMatrix(string transformMatrix)
...
-        /// version of the polygon with that error</param>
+        /// <param name="fix">
+        ///     Invalid polygons always throw an error, this flag determines whether to try to return a repaired
+        ///     version of the polygon with that error
+        /// </param>
         /// <returns>A WKT string with the cleaned polygon</returns>
         public static async Task<string> ValidatePolygonAsync(string wktPolygon, string entityName, bool fix = false)
         {
             // Wrap the private validation method in a Task so we can asynchronously call this non-trivial method
+            // But see the docs: https://docs.microsoft.com/en-us/aspnet/core/performance/performance-best-practices?view=aspnetcore-3.1
```
zmbq, line 81: I'm not sure why they warn against it, as it's not explained. Perhaps they're worried about exhausting the threadpool? Actually, that may be it. You have 20 threads, one of them is handling the current request. You can call the lengthy method on that thread, or you do a Task.Run, free up your thread but make another one block. And you're request is still waiting. All you've done with Task.Run, is add a little overhead. I guess that's the reason.

I'm not sure what the right solution is here. Starting up a Polygon Validation Service that manages its own threads just seem like an overkill here. Perhaps just don't use Task.Run at all, and we'll see if validatePolygon is slowing us down - then we'll worry about it?

https://social.msdn.microsoft.com/Forums/vstudio/en-US/1124354e-7365-430e-9bc4-018138e0d601/creating-multiple-threadpools?forum=netfxbcl - this contains an explanation why lengthy calculations should not be done in thread pool threads.

sqe-database-access/Queries/MutateQueries.cs
```diff
@@ -15,18 +15,24 @@ namespace SQE.DatabaseAccess.Queries
...
-SELECT $Values
+            return @"
+INSERT INTO $TableName ($Columns, creator_id)
+SELECT $Values, @UserId
 FROM dual
 WHERE NOT EXISTS
-  ( SELECT $Columns                 # This is basically an adhoc uniqueness constraint, which Itay wants to protect
-    FROM $TableName                 # against any database schema updates that fail to set a proper uniqueness
-    WHERE ($Columns) {(hasNulls ? "<=>" : "=")} ($Values)    # constraint.  It is very fast if the proper uniqueness constraint already exists.
+  ( SELECT $Columns     # This is basically an adhoc uniqueness constraint, which Itay wants to protect
```
zmbq, line 32: Is # used for database comments? I'm always using -- .

zmbq (approved):

data-to-sqe/qwb-to-sqe/Common/SshDatabase.cs
Bronson-Brown-deVost: I will write an issue on Redmine so we take some time to investigate this.

sqe-api-server/Helpers/InputValidation.cs
Bronson-Brown-deVost: I agree, I think we just remove Task.Run(). From reading the various discussions it is unclear what the best approach is. I understand that one doesn't want a long-running process holding up the thread, but it has to run somewhere and perhaps a dedicated service is the "proper" way to handle it. I guess I will just remove the Task.Run() and leave the comments there.  Thanks for taking a look at it.

sqe-database-access/Queries/MutateQueries.cs
Bronson-Brown-deVost: Yeah, but only for mysql.  I should really switch to --, which should be portable most everywhere.

data-to-sqe/qwb-to-sqe/Common/SshDatabase.cs
Bronson-Brown-deVost: Ah, I didn't look closely enough.  Ingo is using this to connect SQE and QWB for periodic synchronization, so there the cost of setting up the connection is minimal in comparison to the need for a secure connection.  I think this would only run once a day or something like that, I will check with him.

By the way, I don't know why but the main database issue I had was with the naming of a cloned edition directly after creating it.  In the last PR I integrated the renaming directly in the cloning code instead of the original two-step process (clone the edition, then in another transaction rename it), no more deadlock failures.  

I would like to know more about the whole connection system.  Our repos subclass a DbConnectionBase that provides its own OpenConnection(), I see other people also using DI for a transient IDbConnection.  It is a complete mystery to me how any of this performs thread pooling, and nobody seems to be discussing it. After looking at it closely I would think in our setup that each of our repos, which use transient DI, would create a new connection to the database for each individual request. I set up a little log to check and this appears to be the case.  That seems really wasteful, shouldn't we have a long running connection that pulls cursors from a pool?

PR #43, opened 2020-08-05
Custom dto to typescript

Bronson-Brown-deVost:
This is the end result of today's updates to the typescript generation system.


zmbq (approved):
For some reason out of the two Travis tests on the PR one succeeded and the other failed.

PR #44, opened 2020-09-03
Sign interpretation attributes

Bronson-Brown-deVost:
So here is another rather big PR.  Sorry!

You should find in `sqe-api-server-HttpControllers/SignInterpretationController.cs` all the endpoints for dealing with sign interpretations and with sign attributes.  These all have basic tests and are working (so far as the tests know).

I needed to do a bunch of reworking in the test system, which was becoming too cumbersome and error prone.  Now there is a new project GenerateTestRequestObjects, which analyses all the http endpoints and creates a convenient class for the test project to vastly simplify testing http and signalr requests and also running listeners (the generated files all end up in `/sqe-api-tests/ApiRequests` so any comments you might have there I will have to work into the automated parsers/writers of `/Utilities/GenerateTestRequestObjects`).

I tried to increase the code coverage with more tests, specifically targeting the ROI endpoints.  I will have a subsequest PR that tries to finally close the gap between the endpoint functionality and test coverage, which is still not great.

Finally, I moved over to Github Actions, per your suggestion.  Funny enough, while TravisCI failed only intermittently, Github Actions failed every single time, thank you Itay...  After much investigation, soul searching, and anti-witchcraft incantations—all of which made absolutely no positive impact, except for the Akkadian anti-witchraft incantations, of which I am an expert—I discovered that setting the .NET core version to 3.1.201 (instead of 3.1.101) made all the problems disappear (thank you to no one; I could not find this problem or solution anywhere on the web, and yet I too remain silent and perpetuate the problem; how many man-hours must be lost?).  Now the test run properly on Github Actions, and I see no need to switch back to TravisCI.  You will see a step in GithubActions that collects and prints diagnostic data upon failure, I think it is nice to leave this for those occasions when things fail (it should save the need to make secondary runs to investigate the causes of failure).

As you trudge through this PR, I hope a glance back to these initial comments might provide you a moment of levity.


Utilities/GenerateTypescriptInterfaces/Program.cs
```diff
@@ -332,6 +341,12 @@ private class Options
             [Option('h', "hub-folder", Required = false, HelpText = "Specify signalr hub folder path.")]
             public string HubFolder { get; set; }
         }
+
+        private static string StripComments(string comment)
+        {
+            const string re = @"(^|\s+)\/\/(?!\/).*\n";
```
zmbq, line 347: You aren't stripping `/* ... */` comments... I wouldn't worry about it, I don't think it's a problem that comments get into the generated Hub code.

zmbq (approved):

PR #46, opened 2020-10-08
Original connector

Bronson-Brown-deVost:
The PR includes the new sign interpretation attribute and commentary functionality, that has been tested already in the latest frontend development versions.

This also contains substantial updates to the testing system and important fixes to the repositories.

**Repository updates:**
I did not end up changing the database connector, but I did make the following changes:

1. Removed all instances of `await Task.WaitAll()`.  This may have caused trouble by trying to run several async requests to the database at the same time resulting in some unexpected errors and uncertainty with connector management.  In any event, it was simply unnecessary.
2. Made sure that transaction scopes are being handled correctly with possible multithreading and that all transactions include a `Complete()` at the end.
3. Made several edits aimed to make control flow more obvious and error handling more explicit (especially exchanging `FirstOrDefault()` calls for `First()`and explicitly checking for the possibility that a collection was empty).

**Testing Updates**
I increased the coverage so that nearly all endpoints are covered by at least one test.  In order to make test writing simpler, I created a new project under the Utilities folder called `GenerateTestRequestObjects`.  The project examines the HTTP and SignalR endpoints and generates clear well-typed objects that the testing suite can easily use to to test those objects (they are placed in `sqe-api-tests/ApiRequests`.  This has enabled me to greatly increase testing coverage of the SignalR enpoints and also the SignalR listener dispatch system (i.e., notification of other users when a mutation request has been made).  

These updates have resulted in an many more tests being run, and are perhaps responsible for the appearance of more transient errors happening during testing.  After much effort and many hours of work, I decided to try changing the structure of the tests.  Previously, each testing class inherited from a WebFactory class that spun up a fully functioning instance of the API server.  When tests are run, they are processed simultaneously on multiple threads, this meant that four or more independent instances of the server were being run during testing, which likely presented a very large demand on system resources and probably went beyond the actual number of database connections available.  Whether or not that explains the reason for the failures, I rewrote the system so that only one instance of the API server is spun up during testing and after quite a few tests, I have not again run into the errors I had previously experienced.

**Apology**
Sorry that this has turned out to be a massive PR again.  But I figured it best not to let these important updates languish any longer, we need to get these updates out into production sooner rather than later.


sqe-api-server/HttpControllers/RoiController.cs
```diff
@@ -108,7 +108,7 @@ public RoiController(IRoiService roiService, IUserService userService)
         [HttpPut("v1/editions/{editionId}/rois/batch")]
         public async Task<ActionResult<UpdatedInterpretationRoiDTOList>> UpdateRois(
             [FromRoute] uint editionId,
-            [FromBody] InterpretationRoiDTOList updateRois)
+            [FromBody] UpdateInterpretationRoiDTOList updateRois)
```
zmbq, line 111: Hmmmm, this makes me think we've missed something in defining the API on the Typescript side. If I don't change the argument type on the frontend's service, the compiler will not catch this error, and happily pass the wrong type to the server, which will probably fail.

Perhaps we need to automatically generate Typescript methods that call Server endpoints and accept arguments of the types specified in C#?

sqe-api-server/HttpControllers/SignInterpretationController.cs
```diff
@@ -146,13 +152,14 @@ public async Task<ActionResult<AttributeDTO>> CreateEditionSignInterpretationAtt
         }
 
         /// <summary>
-        /// Links two sign interpretations in the edition's sign stream 
+        ///     Links two sign interpretations in the edition's sign stream
```
zmbq, line 155: I really wish there would be a way to tell Github not to diff on things like whitespace, or a line broken into two.

sqe-api-server/RealtimeHubs/RoiHub.cs
```diff
@@ -129,7 +129,7 @@ public async Task<UpdatedInterpretationRoiDTO> PutV1EditionsEditionIdRoisRoiId(u
         /// <param name="editionId">Id of the edition</param>
         /// <param name="updateRois">A JSON object with an array of the updated ROI details</param>
         [Authorize]
-        public async Task<UpdatedInterpretationRoiDTOList> PutV1EditionsEditionIdRoisBatch(uint editionId, InterpretationRoiDTOList updateRois)
+        public async Task<UpdatedInterpretationRoiDTOList> PutV1EditionsEditionIdRoisBatch(uint editionId, UpdateInterpretationRoiDTOList updateRois)
```
zmbq, line 132: Ah, this one will probably cause an error on the Typescript side, because we do have automatically generated methods here.

sqe-api-server/Services/SignInterpretationService.cs
```diff
@@ -329,7 +349,8 @@ await _hubContext.Clients.GroupExcept(user.EditionId.ToString(), clientId)
             return updatedSignInterpretation;
         }
 
-        public async Task<NoContentResult> DeleteSignInterpretationAttributeAsync(UserInfo user, uint signInterpretationId, uint attributeValueId, string clientId = null)
+        public async Task<NoContentResult> DeleteSignInterpretationAttributeAsync(UserInfo user,
+            uint signInterpretationId, uint attributeValueId, string clientId = null)
```
zmbq, line 353: Diff the parsed AST! If github passed each code through the C# parser and compared ASTs, it will only mark changes that are actual changes!

sqe-api-server/Services/TextService.cs
```diff
@@ -342,7 +380,17 @@ private static LineTextDTO _textEditionLineToDTO(TextEdition ed, List<EditorInfo
                                                     attributeString = b.AttributeString,
                                                     attributeValueId = b.AttributeValueId.GetValueOrDefault(),
                                                     attributeValueString = b.AttributeValueString,
-                                                    editorId = b.SignInterpretationAttributeEditorId.GetValueOrDefault(),
+                                                    editorId =
+                                                        b.SignInterpretationAttributeEditorId.GetValueOrDefault(),
+                                                    commentary = string.IsNullOrEmpty(b.AttributeCommentary)
```
zmbq, line 385: Finally a comment on the code itself... This piece of code: commentary = string.IsNullOrEmpty(...) ? null CommentaryDTO { ... } repeats several times, and its quite elaborate. Perhaps placing it in a static function that takes `a` or `b` and returns the right object or null?

sqe-database-access/AttributeRepository.cs
```diff
@@ -211,96 +227,94 @@ public async Task<IEnumerable<SignInterpretationAttributeEntry>> GetEditionAttri
...
-                        x.AttributeStringValue,
-                        x.AttributeStringValueDescription,
-                        x.Css)));
+                        updatedAttributeId,
+                        createAttributeValue.AttributeStringValue,
+                        createAttributeValue.AttributeStringValueDescription,
+                        createAttributeValue.Css);
 
                 // Write the attribute value updates
-                await Task.WhenAll(updateAttributeValues.Select(x =>
```
zmbq, line 254: I would have added an extension method to Task, called WhenAllSeq, that will call `await` in a loop over all Tasks, so all you had to do was change Task.WhenAll to Task.WhenAllSeq everywhere.

sqe-database-access/AttributeRepository.cs
```diff
@@ -211,96 +227,94 @@ public async Task<IEnumerable<SignInterpretationAttributeEntry>> GetEditionAttri
...
-        /// Delete the specified attribute from the edition
+        ///     Delete the specified attribute from the edition
         /// </summary>
         /// <param name="editionUser">The edition user details object</param>
         /// <param name="attributeId">The unique id of the attribute to be deleted</param>
         /// <returns></returns>
         public async Task DeleteEditionAttributeAsync(UserInfo editionUser, uint attributeId)
         {
-            using (var transactionScope = new TransactionScope())
+            using (var transactionScope = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled))
```
zmbq, line 303: I wonder whether this could have been solved by creating an SQEDefaultTransactionScope class or a similar name, and use it throughout the code - so that next time we change something in the transaction behavior, we don't need to change code everywhere.

sqe-database-access/AttributeRepository.cs
```diff
@@ -211,96 +227,94 @@ public async Task<IEnumerable<SignInterpretationAttributeEntry>> GetEditionAttri
...
             {
                 // First get the actual details of the attribute
-                var existingAttribute = await GetEditionAttributeAsync(editionUser, attributeId);
+                var existingAttributes = (await GetEditionAttributeAsync(editionUser, attributeId)).ToList();
```
zmbq, line 306: Why convert to list? Because the underlying cursor may remain open until the last element is retrieved?

sqe-database-access/CatalogueRepository.cs
```diff
@@ -87,11 +87,10 @@ public async Task<IEnumerable<CatalogueMatch>> GetImagedObjectMatchesForTextFrag
             string canonicalEditionName, string canonicalEditionVolume, string canonicalEditionLoc1,
             string canonicalEditionLoc2, byte canonicalEditionSide, string comment)
         {
-            using (var transactionScope = new TransactionScope())
+            using (var transactionScope = new TransactionScope(TransactionScopeAsyncFlowOption.Enabled))
             using (var connection = OpenConnection())
             {
-                //var existingMatches = await GetTextFragmentMatchesForImagedObjectAsync(imagedObjectId);
-                var existingEditionCats = await connection.QueryAsync<EditionCatalogueEntry>(
+                var existingEditionCats = (await connection.QueryAsync<EditionCatalogueEntry>(
```
zmbq, line 93: Wouldn't it be simpler to change QueryAsync to call ToList() before returning the value?

sqe-database-access/DbConnectionBase.cs
```diff
@@ -390,15 +345,15 @@ public override CommandType CommandType
             set => _underlyingSqlCommand.CommandType = value;
         }
 
-        protected override DbConnection DbConnection
+        protected sealed override DbConnection DbConnection
```
zmbq, line 348: Why seal ir? I'm actually not crazy about sealed (or Java's final) at all. I never did understand what it's good for.

sqe-database-access/SignInterpretationRepository.cs
```diff
@@ -39,16 +34,19 @@ public async Task<SignInterpretationData> GetSignInterpretationById(UserInfo use
...
                     _attributeRepository.GetSignInterpretationAttributesByInterpretationId(user, signInterpretationId);
                 var commentaries =
                     _interpretationCommentaryRepository.GetSignInterpretationCommentariesByInterpretationId(user,
                         signInterpretationId);
-                var roiIds = await _roiRepository.GetSignInterpretationRoisIdsByInterpretationId(user, signInterpretationId);
-                var rois = roiIds.Select(x => _roiRepository.GetSignInterpretationRoiByIdAsync(user, x));
+                var roiIds =
+                    await _roiRepository.GetSignInterpretationRoisIdsByInterpretationId(user, signInterpretationId);
+                var rois = new SignInterpretationRoiData[roiIds.Count];
+                foreach (var (roiId, index) in roiIds.Select((x, idx) => (x, idx)))
```
zmbq, line 48: Such loops, in case they pose a performance issue, are prime candidates for a different query - one that returns all ROIs in one query.

zmbq (approved):

sqe-api-server/HttpControllers/RoiController.cs
Bronson-Brown-deVost: I think if you were using the SignalR methods in sqe-signalr.ts for backend calls (instead of HTTP), you would get the integrated safety you are looking for, but I understand that HTTP requests are perhaps easier to debug.  

Very soon, I will integrate an openapi generator (https://github.com/OpenAPITools/openapi-generator), probably into the GitHub actions, so that we generate API libraries.  That could be an option as well, since among other libs I can create a typescript library and push it to NPM for each update.

sqe-api-server/HttpControllers/SignInterpretationController.cs
Bronson-Brown-deVost: Yeah, this frustrates me a lot.  I think if I integrated a more heavy handed "linter", it would fix this.  I do have a "linter" of sorts that runs before every push, but perhaps I can make it even more rigid so we don't get this noise.  My apologies.

sqe-api-server/Services/TextService.cs
Bronson-Brown-deVost: That is a very good idea.  Let me see if I can clean this up in a subsequent PR.  Thanks.

sqe-database-access/AttributeRepository.cs
Bronson-Brown-deVost: Nice idea, I will try that in a subsequent PR.

sqe-database-access/AttributeRepository.cs
Bronson-Brown-deVost: Yeah, there are a couple things like this that could be handy, like a SelectAsync().  I figured the foreach was probably pretty readable though, so I just did that.

sqe-database-access/AttributeRepository.cs
Bronson-Brown-deVost: I did this in a number of places where there was a possible multiple enumeration of an IEnumerable.  And yes, I was a bit paranoid about the possible lazy evaluations, so I hoped this practice would be a bit safer.

sqe-database-access/CatalogueRepository.cs
Bronson-Brown-deVost: I guess so, I just didn't want to create too many abstractions.  I guess I made a bit of a mistake here and should have called Dapper's AsList(), though it probably doesn't matter.

sqe-database-access/DbConnectionBase.cs
Bronson-Brown-deVost: I gave in to the peer pressure of my IDE.  I also hate sealed, it feels simply too bureaucratic. I will fix that in the next PR.

sqe-database-access/SignInterpretationRepository.cs
Bronson-Brown-deVost: Yeah, we should probably add a method that does such a query.  In this case I figured we would usually have only one or two ROI's, so I was lazy and reused existing functionality.  I am putting this as a TODO comment in the next PR.

sqe-database-access/AttributeRepository.cs
Bronson-Brown-deVost: Ha, TransactionScope is sealed, so I had to use a factory instead of simply subclassing it.

PR #48, opened 2020-10-14
Catalogue fixes

Bronson-Brown-deVost:
This is a small PR (especially if you ignore the autogenerated stuff in the sqe-api-test/ApiRequests folder).  It adds the following functionality:

- Bugfix on user creation so that new users now get the proper default system role.
- Added support for the user_data_store in the database with a GET and PUT endpoint for users to add various user settings etc. (do what you will).  The API checks that the submitted data is less than one million characters long and also that it is a valid JSON string.  Let me know if you think we should change either of those—note that the JSON type of the database field probably allows about 1GB, which is way to much to allow for each user, so I wanted the API to restrict that somewhat.
- Some bugfixes in the catalogue endpoints and tests.
- Fixed the error of the license field being NULL in edition DTOs.

### DIFF
--- a/sqe-api-test/UserTests.cs
+++ b/sqe-api-test/UserTests.cs
@@ -228,15 +228,17 @@ namespace SQE.ApiTest
         /// <returns></returns>
         private async Task CleanupUserAccountAsync(UserDTO user)
         {
-            const string deleteNewUserSQL = "DELETE FROM user WHERE email = @Email";
-            const string deleteUserRoleSQL = "DELETE FROM users_system_roles WHERE user_id = @UserID";
-            const string deleteEmailTokenSQL = "DELETE FROM user_email_token WHERE user_id = @UserId";
+            const string deleteNewUserSql = "DELETE FROM user WHERE email = @Email";
+            const string deleteUserRoleSql = "DELETE FROM users_system_roles WHERE user_id = @UserID";
+            const string deleteUserDataSql = "DELETE FROM SQE.user_data_store WHERE user_id = @UserID";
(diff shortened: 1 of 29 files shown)


sqe-api-test/UserTests.cs
```diff
@@ -228,15 +228,17 @@ private async Task<NewUserRequestDTO> CreateRandomActivatedUserAsync(bool should
         /// <returns></returns>
         private async Task CleanupUserAccountAsync(UserDTO user)
         {
-            const string deleteNewUserSQL = "DELETE FROM user WHERE email = @Email";
-            const string deleteUserRoleSQL = "DELETE FROM users_system_roles WHERE user_id = @UserID";
-            const string deleteEmailTokenSQL = "DELETE FROM user_email_token WHERE user_id = @UserId";
+            const string deleteNewUserSql = "DELETE FROM user WHERE email = @Email";
```
zmbq, line 231: I wonder if there's a way to tell Github to ignore specific files in the code review - all of our automatically generated files can be ignored.

sqe-dto/User.cs
```diff
@@ -156,5 +156,13 @@ public class EditorDTO
         public string organization { get; set; }
     }
 
+    public class UserDataStoreDTO
+    {
+        [Required]
+        [StringLength(1000000, MinimumLength = 2,
```
zmbq, line 162: That is more than enough, I do not expect the data to be more than a few KBs. If we want something more substantial, we will probably add a new database entity.

zmbq (approved):

PR #55, opened 2021-01-12
Sign interpretation updates

Bronson-Brown-deVost:
@zmbq, I apologize for this immense PR.  It does include everything up to the latest changes you requested on 11.1.  I also seem to have identified the database "lost connection" problem I mentioned to you.  Apparently Dapper has its own timeout that defaults to 30 seconds, I changed that to 60 seconds (which is the current MariaDB read timeout).  The tests now consistently pass.


sqe-api-server/Serialization/ImagedObjectConversion.cs
```diff
@@ -1,5 +1,6 @@
 using System.Collections.Generic;
 using System.Linq;
+using Dapper;
```
zmbq, line 3: Why add the tiny ORM to a serialization class?

sqe-api-server/Services/ImageService.cs
```diff
@@ -66,7 +69,7 @@ public async Task<InstitutionalImageListDTO> GetInstitutionImagesAsync(string in
 		public async Task<ImagedObjectTextFragmentMatchListDTO> GetImageTextFragmentsAsync(
 				string imagedObjectId)
 		{
-			imagedObjectId = imagedObjectId.Replace("%2F", "/").Replace("%20", " ");
+			imagedObjectId = HttpUtility.UrlDecode(imagedObjectId);
```
zmbq, line 72: Indeed. I wonder how we've missed this one before, decoding just slashes and spaces is risky.

sqe-api-server/Services/SearchService.cs
```diff
@@ -0,0 +1,205 @@
...
+
+			// Find editions first
+			if (!string.IsNullOrEmpty(request.textDesignation))
+			{
+				var editionIds = await _searchRepository.SearchEditions(
+						userId ?? 1
+						, request.textDesignation
+						, request.exactTextDesignation);
+
+				var editionList = await Task.WhenAll(
```
zmbq, line 74: This can be problematic. You get a list of IDs, which is great, but then you issue multiple database transactions - one per each ID, instead of just one transaction (SELECT .... WHERE id IN (id1, id2, id3...))

This will probably not make any difference with one or two users performing searches, but can be an issue when many people search at the same time.

sqe-api-server/Services/SearchService.cs
```diff
@@ -0,0 +1,205 @@
...
+
+			var artefacts = new ArtefactListDTO { artefacts = new List<ArtefactDTO>() };
+
+			var images =
+					new ImageSearchResponseListDTO
+					{
+							imagedObjects = new List<ImageSearchResponseDTO>(),
+					};
+
+			// Find editions first
```
zmbq, line 66: This function is very long. I think it should be split into numerous functions. First one will get the editions, the second one will get the imaged objects based on the editions retrieved by the first one, etc...

sqe-api-server/Services/SearchService.cs
```diff
@@ -0,0 +1,205 @@
...
+
+				editions = new FlatEditionListDTO { editions = editionList.ToDTO() };
+			}
+
+			var searchEditionIds = editions.editions.Select(x => x.id);
+
+			// Find imaged objects
+			if (!string.IsNullOrEmpty(request.imageDesignation))
+			{
+				const string imagedObjectRegex = "(.*)-(.*)-(.*)";
```
zmbq, line 86: I think a few comments explaining which regular expression is relevant in which case would be helpful

sqe-api-server/Services/SignInterpretationService.cs
```diff
@@ -203,70 +228,240 @@ await _hubContext.Clients.GroupExcept(user.EditionId.ToString(), clientId)
...
+					, optional.Contains("delete-all-variants")
+					, false);
 
-			// TODO: Should we also gather and sign interpretations prior to this one in the signstream and
-			// broadcast that information as well? Should that also be returned to the original requester?
+			var deletedList = deleted.ToArray();
+
+			// Prepare the response by gathering created sign interpretation(s) and previous sign interpretations
+			var formattedUpdates = await Task.WhenAll( // Await all async operations
+					updated.Select(
```
zmbq, line 438: This can actually be more problematic than the previous case, as there can be more sign interpretation updates that searches. A lot more. Unless GetEditionSignIntepretation doesn't access the database.

sqe-database-access/Queries/EditionQueries.cs
```diff
@@ -105,7 +105,7 @@ FROM edition_editor
 
     # Get a thumbnail if possible
          LEFT JOIN (
-    SELECT iaa_edition_catalog.manuscript_id, MIN(CONCAT(proxy, url, SQE_image.filename)) AS thumbnail_url
+    SELECT iaa_edition_catalog.manuscript_id, MIN(CONCAT_WS('', proxy, url, SQE_image.filename)) AS thumbnail_url
```
zmbq, line 108: Why use CONCAT_WS here if the separator is an empty string? Any point in not using CONCAT?

sqe-database-access/Queries/EditionQueries.cs
```diff
@@ -162,6 +162,141 @@ internal class Result
 		}
 	}
 
+	internal class EditionQuery
+	{
+		private const string _baseQuery = @"
+SELECT DISTINCTROW ed1.edition_id AS EditionId,
```
zmbq, line 168: I don't know all the MySQL extensions to SQL. DISTINCTROW is another one of those. However, according to the documentation, it is exactly the same as DISTINCT. So why not just use DISTINCT, which everybody understands?

sqe-database-access/Queries/ImageQueries.cs
```diff
@@ -149,8 +148,8 @@ internal class Result
 	internal static class InstitutionImagesQuery
 	{
 		public const string GetQuery = @"
-SELECT image_catalog.object_id AS Name, 
-       CONCAT(image_urls.url, SQE_image.filename, '/full/150,/0/', image_urls.suffix) AS Thumbnail,
+SELECT image_catalog.object_id AS Name,
+       CONCAT_WS('', image_urls.url, SQE_image.filename, '/full/150,/0/', image_urls.suffix) AS Thumbnail,
```
zmbq, line 152: Same question as before- why CONCAT_WS with an empty separator?

sqe-database-access/SearchRepository.cs
```diff
@@ -0,0 +1,197 @@
...
+WHERE manuscript_data.name $Match
+    AND (edition.public = 1 OR edition_editor.user_id = @UserId)
+LIMIT 100
+";
+
+				sql = sql.Replace(
+						"$Match"
+						, exact
+								? "= @ManuscriptName"
+								: "LIKE CONCAT('%', @ManuscriptName, '%')");
```
zmbq, line 58: Ah, I wondered how 'exact' was implemented (or rather, how inexact was implemented). We should think about ordering results based on the distance of the inexact match from the result. So that an exact match will appear before inexact ones.

sqe-database-access/SignInterpretationRepository.cs
```diff
@@ -132,5 +145,99 @@ UserInfo user
...
+		public async Task UpdateSignInterpretationCharacterById(
+				UserInfo user
+				, uint   signInterpretationId
+				, string character
+				, byte   priority
+				, uint   attributeValueId)
+		{
+			using (var transactionScope =
+					new TransactionScope(TransactionScopeAsyncFlowOption.Enabled))
+			using (var conn = OpenConnection())
```
zmbq, line 158: We should most definitely figure out a way to not open multiple connections per request, this really messes things up with locking. We need a very good performance reason to have more than one connection per request. I'm not sure how we're going to do that, but we really should.

sqe-database-access/TextRepository.cs
```diff
@@ -325,10 +366,243 @@ UserInfo editionUser
 			return new LineData { LineId = lineId, LineName = lineName };
 		}
 
+		/// <summary>
+		///  Inserts new line as fitstline into an existing text fragment which masu contasin
```
zmbq, line 370: There are some typos in this comment....

sqe-database-access/TextRepository.cs
```diff
@@ -325,10 +366,243 @@ UserInfo editionUser
...
+								, newLine.Signs.Last()
+										 .SignInterpretations[0]
+										 .SignInterpretationId.Value
+								, attr);
+
+				// Add the attribute also to the return value.
+				newLine.Signs.Last().SignInterpretations[0].Attributes.AddRange(newAttributes);
+
+				// Delete the end break from the old last line
+				//TODO How do we signal this change?
```
zmbq, line 515: Perhaps in case of changes that are substantial (here, in the previous function and maybe later as well), we should have a catch all "please query this again" SignalR message, so clients know their sign interpretation data is invalid, and they need to query it again?)

sqe-database-access/TextRepository.cs
```diff
@@ -354,43 +628,251 @@ UserInfo editionUser
...
+						, IEnumerable<SignData> newSigns
+						, List<uint>            anchorsBefore
+						, List<uint>            anchorsAfter
+						, uint?                 signInterpretationId
+						, bool                  breakNeighboringAnchors = false
+						, bool                  materializeSignStream   = true)
 		{
-			var newSigns = new List<SignData>();
+			// TODO: This method is a total mess, nesting should be reduces by using subroutines
+			// there are a lot of conditionals that are too hard to follow, and I doubt it will
```
zmbq, line 649: Indeed, it is very hard to follow and understand.

sqe-database-access/TextRepository.cs
```diff
@@ -424,13 +906,36 @@ UserInfo                editionUser
...
 
-			return newSigns;
+			// Rebuild any affected materialized sign streams
+			// Do not await it, we don't care if it finishes before this
+			// request is over.
+			// TODO: change RequestMaterializationAsync to take an array,
+			// it should be able to check how many sign streams would need a rebuild.
+			if (materializeSignStream && updatedSignInterpretationIds.Any())
+			{
+				_materializationRepository.RequestMaterializationAsync(
```
zmbq, line 933: I am not sure if we really want to leave something running after the request is done - in this process. This may be OK for the short term, but really if this is a lengthy operation, we should probably off-load it to a different process.

sqe-database-access/TextRepository.cs
```diff
@@ -446,39 +951,69 @@ UserInfo                editionUser
...
+				, IEnumerable<uint> firstSignInterpretationIds
+				, IEnumerable<uint> secondSignInterpretationIds
+				, bool              materializeSignStream = true)
 		{
+			// First check against cycles in the graph
+			foreach (var firstSI in firstSignInterpretationIds)
+			{
+				foreach (var secondSI in secondSignInterpretationIds)
+				{
+					var createsCycle = await _materializationRepository.IsCycleAsync(
```
zmbq, line 1000: I have this half-baked idea of imposing a total order on the partial order of signs we have, which will make cycle detection extremely easy. I'm not sure it's possible in our case, Let's discuss this on Monday.

zmbq (approved):
Wow, this has been a big one!

I think some of the comments I've made are important and need to be addressed, but not urgently - so I'm approving this PR.

sqe-api-server/Serialization/ImagedObjectConversion.cs
Bronson-Brown-deVost: Premature optimization I think.  Dapper offers an AsList() function that converts its enumerated results to a list in place without any copying.  Probably this is pointless.

sqe-api-server/Services/ImageService.cs
Bronson-Brown-deVost: Yeah, please keep your eyes out for that, the original code was pretty dangerous.  I did a search in the project for %2 and I got no results, so hopefully I did not make this mistake anywhere else.

sqe-api-server/Services/SearchService.cs
Bronson-Brown-deVost: I put in a TODO note.  This should indeed be fixed to support one DB query with a list of ids.

sqe-api-server/Services/SearchService.cs
Bronson-Brown-deVost: Agreed.  I put a note in for this too.

sqe-api-server/Services/SearchService.cs
Bronson-Brown-deVost: I added the comments and wrote a note that we will need to revisit the string parsing logic once this is used a little bit by real people.

sqe-database-access/TextRepository.cs
Bronson-Brown-deVost: Thanks.

sqe-database-access/TextRepository.cs
Bronson-Brown-deVost: Yeah, this method is not yet exposed in the API, but we will need to do something like that.

sqe-database-access/TextRepository.cs
Bronson-Brown-deVost: Agreed, what are your suggestions for doing that?  Do we use BackgroundService?

